PR #1, opened 2020-01-26
Signup APIs

mittalyashu:
Creating sign up API at `/api/v1/auth/signup` route.


server/controllers/auth/signup.js
```diff
@@ -0,0 +1,83 @@
+// modules 
+const bcrypt = require('bcrypt');
+const uuid = require('uuid/v1');
+const jwt = require('jsonwebtoken');
```
tejas-hosamani, line 4: If you plan on providing login access via social media accounts like, Facebook, GitHub, Google, consider using `Passport` instead of `jsonwebtoken`. Because it comes with all those plus jwt.

server/controllers/auth/signup.js
```diff
@@ -0,0 +1,83 @@
...
+
+const database = require("../../database");
+
+exports.signup = (req, res, next) => {
+  const email = req.body.email;
+  // password cannot be longer than 72 characters
+  const password = req.body.password;
+
+  // validate member existance in database
+  database.query(`
```
tejas-hosamani, line 14: Maybe use tools such as https://knexjs.org/ for query building? It comes with migration options as well. Not sure how you are managing migrations right now.

server/controllers/auth/signup.js
```diff
@@ -0,0 +1,83 @@
...
+      database
+        .query(`
+          INSERT INTO
+            member(member_id, email_address, password)
+          VALUES
+            ('${memberId}', '${email}', '${passwordHash}')
+          RETURNING
+            member_id, email_address, created_at;
+        `)
+        .then(result => {
```
tejas-hosamani, line 57: `await` would improve code readability.

server/controllers/auth/signup.js
mittalyashu: > If you plan on providing login access via social media accounts like, Facebook, GitHub, Google

Well, that's not the plan thought.

But I want the website owner can allow their user to authenticate from their site and re-direct to this site, something like SSO.

server/controllers/auth/signup.js
mittalyashu: I am also not sure about the migration, I haven't look much into that.

But I did thought of using sequelize too, but using ORMs doesn't provide much flexibility as compare to native SQL language.

server/controllers/auth/signup.js
mittalyashu: Can you share any small example?

I did tried to break the code into multiple lines to make it readable.

server/controllers/auth/signup.js
tejas-hosamani: > but using ORMs doesn't provide much flexibility as compare to native SQL language.

Technically true. But haven't come across any query that I couldn't write with ORM. Plus, you can always use regular SQL if such instance occurs. Overall, ORMs provide better readability.

Plus, the migration is an important factor. If it were a MongoDB, it would create the DB "schema" on its own. But SQL DBs don't. If multiple people were to work on this project, setting-up project locally would become a hassle.

server/controllers/auth/signup.js
tejas-hosamani: Something like:
```js
const insertedUser = await DB.query(query);
const authToken = jwt.sign(insertedUser, 'scretKey', { expiresIn: '1h' });
```

mittalyashu: I am trying to keep the code consistent by using promises everywhere.

Using `async/await` at one place will make it in-consistent.

tejas-hosamani: Not at one place. I suggest to use async await everywhere. 

Let's keep going for now until first release. We could re-think these later.

PR #4, opened 2020-02-01
Login API

mittalyashu:


server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+      if (validatePassword) {
+
+        const loginToken = jwt.sign({
+          member_id: memberData.member_id,
+          email_address: memberData.email_address,
+          is_owner: memberData.is_owner,
+          is_moderator: memberData.is_moderator,
+          is_blocked: memberData.is_blocked,
+          created_at: memberData.created_at
+        }, 'secretKey', { expiresIn: '2d' });
```
tejas-hosamani, line 35: It's better to reference secrete key from `.env` file rather than hard coding it.

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+            email_address: memberData.email_address,
+            profile_picture: memberData.profile_picture,
+            is_owner: memberData.is_owner,
+            is_moderator: memberData.is_moderator,
+            is_blocked: memberData.is_blocked,
+            created_at: memberData.created_at
+          },
+          token: loginToken
+        })
+        next();
```
tejas-hosamani: `next()` should never occur after response has dispatched. So, this needs to be removed.

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+          status: {
+            code: 409,
+            type: "error"
+          },
+          error: {
+            code: "invalid_password",
+            message: "Password is incorrect"
+          }
+        })
+        next();
```
tejas-hosamani: Same with this `next()`

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+      status: {
+        code: 409,
+        type: "error"
+      },
+      error: {
+        code: "member_not_found",
+        message: "Member not found."
+      }
+    })
+    next();
```
tejas-hosamani: Same here as well.

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
+// modules
+const bcrypt = require('bcrypt');
+const jwt = require('jsonwebtoken');
+
+const database = require("../../database");
+
+exports.login = (req, res, next) => {
+  // todo: 'remember me' feature
+  const email = req.body.email;
+  const password = req.body.password;
```
tejas-hosamani, line 10: You could de-structure them whenever possible.

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+    FROM
+      member
+    WHERE
+      email_address = '${email}'
+  ;`).then(databaseResult => {
+    // get first member from array of objects
+    const memberData = databaseResult.rows[0];
+
+    // validate member password with encrypted password in database
+    bcrypt.compare(password, memberData.password).then(validatePassword => {
```
tejas-hosamani, line 25: This has two level nested `then`. `async/await` would be better here.

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+      if (validatePassword) {
+
+        const loginToken = jwt.sign({
+          member_id: memberData.member_id,
+          email_address: memberData.email_address,
+          is_owner: memberData.is_owner,
+          is_moderator: memberData.is_moderator,
+          is_blocked: memberData.is_blocked,
+          created_at: memberData.created_at
```
tejas-hosamani, line 34: Is `created_at` necessary in token sign?

server/controllers/auth/login.js
```diff
@@ -0,0 +1,85 @@
...
+          error: {
+            code: "invalid_password",
+            message: "Password is incorrect"
+          }
+        })
+        next();
+      }
+    });
+  }).catch(error => {
+    console.log(error);
```
tejas-hosamani: Make this `console.error` instead?

server/controllers/auth/login.js
tejas-hosamani: Also, you could implement Joi validation here by taking reference from my PR? I think it has same exact properties.

server/controllers/auth/login.js
mittalyashu: On that topic, I did posted on site for getting thoughts from other developers about it.

https://hashnode.com/post/can-we-use-user-password-as-jwt-secret-key-ck5vzwvmv0894qks1978ib9gh

To create or use a better secret key.

server/controllers/auth/login.js
mittalyashu: ok

server/controllers/auth/login.js
mittalyashu: ok

server/controllers/auth/login.js
mittalyashu: ok

server/controllers/auth/login.js
mittalyashu: > You could de-structure them whenever possible.

What do you mean by de-structure?

> Also, you could implement Joi validation here by taking reference from my PR? I think it has same exact properties.

Will do that once this PR is merged.

server/controllers/auth/login.js
mittalyashu: Initially when I was working on the token payload, there was very less data for me to create the actual token, so I threw `created_at` 😅.

server/controllers/auth/login.js
mittalyashu: ok

server/controllers/auth/login.js
mittalyashu: 🤔

server/controllers/auth/login.js
tejas-hosamani: Dude, that's an awesome idea! I never considered that! 😆

mittalyashu:
@CodingCreate101 Can you use the PR suggest changes feature can suggest those changes directly.

I am working on something else locally, so I cannot checkout to other branch right now.

server/controllers/auth/login.js
mittalyashu: _yeah, I still need to reply to their comments, so let's keep it for now._

Let's hear what other devs have to say and then we can make changes accordingly.

server/controllers/auth/login.js
tejas-hosamani: ` const { email, password } = req.body;` 
something like that.

server/controllers/auth/login.js
tejas-hosamani: lol. Put user firstName maybe, if that's still the case?

tejas-hosamani, line 34:
```suggestion
          first_name: memberData.first_name,
```

server/controllers/auth/login.js
tejas-hosamani:
```suggestion
       
```

server/controllers/auth/login.js
tejas-hosamani:
```suggestion
       
```

server/controllers/auth/login.js
tejas-hosamani:
```suggestion
    console.error(error);
```

server/controllers/auth/login.js
tejas-hosamani:
```suggestion
    
```

server/controllers/auth/login.js
mittalyashu: > lol. Put user firstName maybe, if that's still the case?

Those are all null values. Didn’t you saw the database on testing the signup API?

tejas-hosamani: Oh, crap. You are right.

server/controllers/auth/login.js
mittalyashu: Keeping it expanded makes the code readable.

mittalyashu:
Login API looks good to me. Merging this PR.

PR #6, opened 2020-02-01
Bash scripts for database setup

mittalyashu:
Creating bash scripts for setting up postgreSQL database locally.

### DIFF
--- a/scripts/database-setup.sh
+++ b/scripts/database-setup.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+
+# environment variables
+source ./scripts/variables.sh
+
+echo
+echo "Creating database table"
+echo
+
+echo "==== member table ===="
+psql \
+-c \
+  "CREATE TABLE IF NOT EXISTS member (
+    member_id VARCHAR(80) PRIMARY KEY,
+    first_name VARCHAR(30),
+    last_name VARCHAR(30),
+    email_address VARCHAR(320) UNIQUE NOT NULL,
+    password VARCHAR(72) NOT NULL,
+    profile_picture TEXT,
+    is_verified BOOLEAN DEFAULT false,
+    is_owner BOOLEAN DEFAULT false,
+    is_moderator BOOLEAN DEFAULT false,
+    is_blocked BOOLEAN DEFAULT false,
+    created_at timestamp DEFAULT current_timestamp
+  );" \
+-d $PG_DATABASE \
+-h $PG_HOST \
+-p $PG_PORT \
+-U $PG_USER
+echo
+
+echo "==== post table ===="
+psql \
+-c \
+  "CREATE TABLE IF NOT EXISTS post (
+    post_id VARCHAR(80) PRIMARY KEY,
+    title VARCHAR(100) NOT NULL,
+    slug VARCHAR(150) NOT NULL,
+    body_markdown TEXT,
+    member_id VARCHAR(80) NOT NULL REFERENCES member(member_id),
+    category_id VARCHAR(80) NOT NULL REFERENCES category(category_id),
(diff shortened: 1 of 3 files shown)


scripts/database-setup.sh
```diff
@@ -0,0 +1,137 @@
...
+    is_verified BOOLEAN DEFAULT false,
+    is_owner BOOLEAN DEFAULT false,
+    is_moderator BOOLEAN DEFAULT false,
+    is_blocked BOOLEAN DEFAULT false,
+    created_at timestamp DEFAULT current_timestamp
+  );" \
+-d $PG_DATABASE \
+-h $PG_HOST \
+-p $PG_PORT \
+-U $PG_USER
```
mittalyashu, line 29: @CodingCreate101 any way to avoid writing these 4 lines again and again.

scripts/database-setup.sh
```diff
@@ -0,0 +1,137 @@
...
+#     (status_id, name, slug)
+# VALUES
+#   (1, 'Under Review', 'under-review'),
+#   (2, 'Planned', 'planned'),
+#   (3, 'In Progress', 'in-progress')
+#   );" \
+# -d $PG_DATABASE \
+# -h $PG_HOST \
+# -p $PG_PORT \
+# -U $PG_USER
```
mittalyashu: @CodingCreate101 Is it good practice to write insert query here for real?

tejas-hosamani: I think these should be part of different files entirely, as initial seeds. But for the sake of MVP, let's move on. I ll just use this to setup right now. I ll update if possible later this week, one we reach our this week's release goal.

scripts/database-setup.sh
tejas-hosamani: It could be used as a variable.
Ex:
`CONFIG=-d $PG_DATABASE -h $PG_HOST -p $PG_PORT -U $PG_USER`

and use `$CONFIG` wherever we want.

Not sure if the syntax is accurate but something like that could be possible. You might have already tried this though.

scripts/database-setup.sh
```diff
@@ -0,0 +1,137 @@
...
+# psql \
+# -c \
+#   "INSERT INTO
+#     category
+#       (category_id, name, slug, hex_color)
+#   VALUES
+#     (1, 'New', 'new', '3778ff'),
+#     (2, 'Improvement', 'improvement', '11C63C'),
+#     (3, 'Fix', 'fix', 'ff4772')
+#   );" \
```
tejas-hosamani: Had to remove this parenthesis `)` to make it work.

Same with next `INSERT` instruction as well. Had to remove the last `)`.

scripts/database-setup.sh
mittalyashu: Hmm...

I tried to convert this part into a function and then use it, but `psql` command was converting the name of the function as database name.

I can try with the code you have mentioned, like converting the code into the variable itself.

scripts/database-setup.sh
mittalyashu: > I think these should be part of different files entirely, as initial seeds.

I think there should be a lot of default data in each table so that the super member (site admin) can actually understand like what's happening, what do you think?

Of course, the super admin can remove those default data too.

> let's move on. I ll just use this to setup right now. I ll update if possible later this week, one we reach our this week's release goal.

👍

scripts/database-setup.sh
mittalyashu:
```suggestion
#   ;" \
```

You mean this one?

tejas-hosamani: Yes. Because there is no opening pair of it, yeah?

PR #9, opened 2020-02-02
Added bcrypt + few additional settings for DB config.

CodingCreate101:


package.json
```diff
@@ -10,7 +10,7 @@
     "server-start": "node -r dotenv/config server/app.js"
   },
   "dependencies": {
-    "bcrypt": "^3.0.7",
+    "bcryptjs": "^2.4.3",
```
tejas-hosamani, line 13: This package hasn't been updated since 3 years. So, we need to look out for `bcrypt` installation fix and add it back again once it starts working.

server/app.js
```diff
@@ -8,6 +8,8 @@ const routes = require("./routes/v1");
 
 const app = express()
 
+const SERVER_PORT = process.env.SERVER_PORT;
```
tejas-hosamani: Added separate variable because, eventually I think we need to validate these env vars based on environment that the application is going to run in. Ex: development, staging, production etc.

package.json
mittalyashu: Ok

server/app.js
mittalyashu: I think we don't need this, if we're gonna merge this PR #6

mittalyashu:
Can you revert the changes of `readme.md` file.

I will later put down a proper documentation just [like this one](https://github.com/gitthermal/thermal/blob/master/README.md).

---

To revert back the changes, you can either use `git revert` command for this one `git checkout <commit_id> -- file1/to/restore file2/to/restore`.

Let me know if you need help with reverting back the commit.

PR #11, opened 2020-02-02
Auth modal (login & signup)

mittalyashu:


mittalyashu:
@CodingCreate101 Could you please review this PR as well.

mittalyashu:
In this PR, I have not imported that auth modal anywhere, where we should place that modal so that it can be easily trigged?

tejas-hosamani:
Sure, I'll pull this and review it, that would be easier. Probably today evening.

PR #12, opened 2020-02-02
Added validation for signup request

CodingCreate101:


server/middlewares/index.js
```diff
@@ -0,0 +1,26 @@
...
+			message,
+		},
+    stack: process.env.NODE_ENV === 'production' ? '🥞' : stack,
+  });
+}
+
+module.exports = {
+  errorHandler,
+  notFound,
+};
```
mittalyashu: What is the use-case of this whole middleware?

server/controllers/auth/signup.js
```diff
@@ -1,14 +1,34 @@
...
 exports.signup = (req, res, next) => {
-  const email = req.body.email;
-  // password cannot be longer than 72 characters
-  const password = req.body.password;
+
+	const valid = signupSchema.validate(req.body);
+
+	if (valid.error) {
+		res.status(422);
+		next(valid.error);
```
mittalyashu, line 27: Why do we need `next()` after sending the response?

server/middlewares/index.js
tejas-hosamani: When we send errors via `next` method, those will be handled by these middledwares.

server/controllers/auth/signup.js
tejas-hosamani: `res.status(422);` is to set the status. `next` actually handles the error response via middlewares.

mittalyashu:
I am also getting this error

```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:470:11)
    at ServerResponse.header (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/response.js:170:12)
    at ServerResponse.json (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/response.js:267:15)
    at errorHandler (/Users/yashumittal/code/logchimp/logchimp/server/middlewares/index.js:11:22)
    at Layer.handle_error (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/router/layer.js:71:5)
    at trim_prefix (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/router/index.js:315:13)
    at /Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/router/index.js:335:12)
    at next (/Users/yashumittal/code/logchimp/logchimp/node_modules/express/lib/router/index.js:275:10)
```

it did return the data from database after saving the data.

server/controllers/auth/signup.js
mittalyashu: I think it is giving some kind of error https://github.com/logchimp/logchimp/pull/12#issuecomment-581121068

tejas-hosamani:
That's because there is `next()` after response has already dispatched somewhere. For which route are you getting this?

mittalyashu:
> That's because there is `next()` after response has already dispatched somewhere. For which route are you getting this?

`/auth/signup`

tejas-hosamani:
That's ~probably~ because of: https://github.com/logchimp/logchimp/pull/12/files#diff-f370aa8c7f2cec86df97ece87fa408bbL37

I've already removed it.

server/controllers/auth/signup.js
tejas-hosamani: No no, that error is because there is next after response dispatch. `res.status(422);` doesn't dispatch response. It only updates the response status.

server/middlewares/index.js
mittalyashu: But where are you using it exactly?

server/controllers/auth/signup.js
```diff
@@ -1,14 +1,34 @@
...
+		return false;
+	}
+
+	const { email, password } = valid.value;
```
mittalyashu, line 31:
```suggestion
	const email = valid.value.email;
	const password = valid.value.password;
```

Can you expand it? For better readability.

server/middlewares/index.js
tejas-hosamani: Here: https://github.com/logchimp/logchimp/pull/12/files#diff-9e4a215b4f5f613a802fa43d8ab35270R33

server/controllers/auth/signup.js
tejas-hosamani: de-structuring is considered a cleaner way. These are only two values, but in case we have more, it would make code lengthy unnecessarily.

server/controllers/auth/signup.js
```diff
@@ -1,14 +1,34 @@
...
+
 exports.signup = (req, res, next) => {
-  const email = req.body.email;
-  // password cannot be longer than 72 characters
-  const password = req.body.password;
+
+	const valid = signupSchema.validate(req.body);
+
+	if (valid.error) {
+		res.status(422);
```
mittalyashu, line 26: You are just sending the status code but not sending any response to the client?

tejas-hosamani: The way it works is, the `res` object at this line has status code. The next line has the `next()` method with error. Since there is no other middleware after this one, the code execution continues and reaches error handler middleware, and that's where the response is handled with this error code.

server/middlewares/index.js
mittalyashu: I still don't get it, I guess that's fine.

Can you format the middleware according to this? https://github.com/logchimp/logchimp/pull/13/files

server/controllers/auth/signup.js
mittalyashu: 🤔in that case... ok

server/controllers/auth/signup.js
mittalyashu: I haven't seen that method to pass error to middleware further.

tejas-hosamani: `app.use()` is intended to use in specific order. Error handlers are placed at the very end for this reason. When no other middlewares respond, the error middleware takes over the response.

tejas-hosamani:
It's better to squash all the merges. Don't you think? @mittalyashu

mittalyashu:
Yes, I did squash the commits while merging.

PR #13, opened 2020-02-02
Token validation middleware

mittalyashu:
Getting the token from the request headers and using it for validation.

### DIFF
--- a/server/middlewares/token.js
+++ b/server/middlewares/token.js
@@ -1,0 +1,46 @@
+// modules
+const jwt = require('jsonwebtoken');
+
+exports.validate = (req, res, next) => {
+  const token = req.headers['authorization'];
+
+  if (token) {
+    // grab token from headers request
+    const headersToken = token.split(" ")[1];
+
+    jwt.verify(headersToken, "secretKey", err => {
+      if (err) {
+        console.log(err);
+
+        /**
+         * sending response to client
+         * only on if the token is missing
+         */
+        res.status(401).send({
+          status: {
+            code: 401,
+            type: "error"
+          },
+          error: {
+            code: "token_invalid",
+            message: "Token is invalid"
+          }
+        })
+      } else {
+        // move to next middleware on success
+        next();
+      }
+    })
+  } else {
+    res.status(401).send({
+      status: {
+        code: 401,
+        type: "error"
+      },
+      error: {
+        code: "token_missing",
+        message: "Token is missing"
+      }
+    })
+  }
+}
--- a/server/middlewares/token.js
+++ b/server/middlewares/token.js
@@ -10,7 +10,7 @@ exports.validate = (req, res, next) => {
 
     jwt.verify(headersToken, "secretKey", err => {
       if (err) {
-        console.log(err);
+        console.error(err);
 
         /**
          * sending response to client


server/middlewares/token.js
```diff
@@ -0,0 +1,46 @@
...
+exports.validate = (req, res, next) => {
+  const token = req.headers['authorization'];
+
+  if (token) {
+    // grab token from headers request
+    const headersToken = token.split(" ")[1];
+
+    jwt.verify(headersToken, "secretKey", err => {
+      if (err) {
+        console.log(err);
```
tejas-hosamani: This could be `console.error`

server/middlewares/token.js
```diff
@@ -0,0 +1,46 @@
...
+const jwt = require('jsonwebtoken');
+
+exports.validate = (req, res, next) => {
+  const token = req.headers['authorization'];
+
+  if (token) {
+    // grab token from headers request
+    const headersToken = token.split(" ")[1];
+
+    jwt.verify(headersToken, "secretKey", err => {
```
tejas-hosamani, line 11: I was thinking about `secretKey` yesterday, using user's password.

You'd have to query users' password every time to validate token. That'd basically make every query expensive.

Also as people on Hashnode mentioned, users' passwords aren't secure anyway.

If you want to keep secretekey different for everyone, you could generate a random key and store that in DB. And use that to sign the token and to verify. This can be different for everyone.

server/middlewares/token.js
```diff
@@ -0,0 +1,46 @@
...
+        })
+      } else {
+        // move to next middleware on success
+        next();
+      }
+    })
+  } else {
+    res.status(401).send({
+      status: {
+        code: 401,
```
tejas-hosamani, line 37: If the token is missing, shouldn't it be 400/422?

mittalyashu: https://httpstatuses.com/401

I am using 401 because it means unauthorised, which means member is trying to access something with missing token.

server/middlewares/token.js
mittalyashu: > If you want to keep `secretekey` different for everyone, you could generate a random key and store that in DB. And use that to sign the token and to verify. This can be different for everyone.

How about using email as secret key? Since each account as a unique email address.

tejas-hosamani: What? No, email ID is public data. Anybody who knows your email ID, has your token forever, that's what it means.

mittalyashu: Hmm... 🤔

mittalyashu: How about.. We get the timestamp at the time of member login and store inside `member` table _(maybe that data can be later helpful for creating analytics dashboard for admins, anyway)_.

Then, at the time of token validation we can get the `login_request_timestamp` _(name of the database column)_ and decode the token.

How this sounds? It follows what you mentioned about generating a random key and storing inside the database.

tejas-hosamani: But timestamp isn't random though. Since this is an open source project, anybody can look at the code and see that it's using timestamp, Since we are expiring token every 2days, user has to login again every 2 days, what that means is, someone can generate token for every second for 2 day period and try using them all as part of brute force method. This isn't hard to do programmatically, and it would hardly take an hour or even less. And if that someone can guess the login time roughly, say maybe he/she knows someone logged in couple of hour ago, brute force method would yield results even quicker.

mittalyashu: So, we should use UUID to generate a `secretKey`?

mittalyashu:
@CodingCreate101 I am gonna merge this PR.

Changes regarding the secret key in can done in separate PR.

PR #14, opened 2020-02-03
posts APIs

mittalyashu:


mittalyashu:
@CodingCreate101 Please review this PR too.

server/controllers/post/create.js
```diff
@@ -0,0 +1,52 @@
...
+
+const database = require('../../database')
+
+exports.create = (req, res, next) => {
+	const postTitle = req.body.title;
+	const postSlug = req.body.slug;
+	const bodyMarkdown = req.body.bodyMarkdown;
+	const memberId = req.body.memberId;
+	const categoryId = req.body.categoryId;
+	const statusId = req.body.statusId;
```
tejas-hosamani, line 12: Sure to not use de-structuring?

server/controllers/post/create.js
```diff
@@ -0,0 +1,52 @@
+// modules
+const uuid = require('uuid/v1');
+
+const database = require('../../database')
+
+exports.create = (req, res, next) => {
+	const postTitle = req.body.title;
+	const postSlug = req.body.slug;
```
tejas-hosamani, line 8: Generating post slug from frontend using post title?

Wouldn't it be better to create it at the backend itself? Because, we shouldn't be sending unnecessary data inside requests. It's not much data but still.

server/controllers/post/create.js
```diff
@@ -0,0 +1,52 @@
...
+				code: 200,
+				type: "success"
+			},
+			post: {
+				postId: postData.post_id,
+				slug: postData.slug
+			}
+		});
+	}).catch(error => {
+		console.log(error);
```
tejas-hosamani: `console.error(error);`

Also, respond back error to user as well?

server/controllers/post/create.js
```diff
@@ -0,0 +1,52 @@
...
+	const categoryId = req.body.categoryId;
+	const statusId = req.body.statusId;
+
+	// generate unique indentification
+	const postId = uuid()
+
+	const postSlugId = `${postSlug}-${Date.now().toString(36)}`
+
+	database.query(`
+    INSERT INTO
```
tejas-hosamani, line 20: Post data should be validated before inserting.

server/controllers/post/deleteById.js
```diff
@@ -0,0 +1,20 @@
+const database = require('../../database');
+
+exports.deleteById = (req, res, next) => {
+	const postId = req.query.postId;
+
+	database.query(`
+    DELETE FROM post
```
tejas-hosamani, line 7: Shouldn't you be checking if the post exists before deleting?

server/controllers/post/deleteById.js
```diff
@@ -0,0 +1,20 @@
...
+  ;`).then(post => {
+		// todo: improve the response output
+		res.status(200).send({
+			status: {
+				code: 200,
+				type: "success"
+			}
+		})
+	}).catch(error => {
+		console.log(error);
```
tejas-hosamani: `console.error(error);`

server/controllers/post/getPostById.js
```diff
@@ -0,0 +1,36 @@
...
+					type: "error"
+				},
+				error: {
+					code: "post_not_found",
+					message: "Post not found."
+				}
+			})
+		}
+	}).catch(error => {
+		console.log(error);
```
tejas-hosamani: `console.error(error);`

server/controllers/post/index.js
```diff
@@ -0,0 +1,11 @@
+const create = require('./create');
+const getPostById = require('./getPostById');
+const updatePostById = require('./updatePostById');
+const deleteById = require('./deleteById');
+
+module.exports = {
+	...create,
+	...getPostById,
+	...updatePostById,
+	...deleteById
```
tejas-hosamani, line 12: Why are you spreading them?

server/controllers/post/updatePostById.js
```diff
@@ -0,0 +1,40 @@
...
+	const bodyMarkdown = req.body.bodyMarkdown;
+	const categoryId = req.body.categoryId;
+	const statusId = req.body.statusId;
+
+  /**
+   * note: do not add memberId and postId
+   * inside this query
+   */
+	database.query(`
+    UPDATE post
```
tejas-hosamani, line 16: Check whether post exists before updating. Plus, data validation is necessary too.

server/controllers/post/updatePostById.js
```diff
@@ -0,0 +1,40 @@
...
+    SET
+      title = '${postTitle}',
+      slug = '${postSlug}',
+      body_markdown = '${bodyMarkdown}',
+      category_id = '${categoryId}',
+      status_id = '${statusId}',
+      updated_at = current_timestamp
```
tejas-hosamani, line 23: Is `current_timestamp` inbuilt variable or something?

server/controllers/post/updatePostById.js
```diff
@@ -0,0 +1,40 @@
...
+
+		res.status(200).send({
+			status: {
+				code: 200,
+				type: "success"
+			},
+			post: postData
+		})
+	}).catch(error => {
+		console.log(error);
```
tejas-hosamani: `console.error(error);`

Also, return error to user as well?

server/routes/v1/posts.js
```diff
@@ -0,0 +1,16 @@
...
+const router = express.Router();
+
+// controller
+const post = require('../../controllers/post');
+
+// middleware
+const token = require('../../middlewares/token');
+
+router.post('/post/create', token.validate, post.create)
+router.get('/post/:slug', post.getPostById)
```
tejas-hosamani, line 13: User should be logged in to view posts right? We need token validation here as well, no?

server/controllers/post/create.js
mittalyashu: I know that using destructing methods can be done in few lines of code, but it doesn't support the fact that if un-even number of data is passes, in that case the name and its value might get shuffled.

server/controllers/post/create.js
mittalyashu: I still need to write the logic to convert title to slug.

Any thoughts on that?

server/controllers/post/create.js
mittalyashu: > Also, respond back error to user as well?

What kind of error respond to the user?

server/controllers/post/create.js
mittalyashu: I am not sure which fields are required, so I will delay this until I work on the frontend.

server/controllers/post/deleteById.js
mittalyashu: Well, based on my understand this feature (a.k.a feature) will only be accessible after opening the post itself.

server/controllers/post/deleteById.js
mittalyashu: ok

server/controllers/post/getPostById.js
mittalyashu: ok

server/controllers/post/index.js
mittalyashu: On requiring each module well, they are accessible using `create.create` like object inside an object so to avoid doing that I used the spread operator.

server/controllers/post/updatePostById.js
mittalyashu: Yes

server/controllers/post/create.js
tejas-hosamani: > un-even number of data

What do you mean?

server/controllers/post/create.js
tejas-hosamani: First thoughts:
1. Make all chars lowercase
1. Replace space with `-`
1. Append the random ID

Seems enough, I think.

server/controllers/post/create.js
tejas-hosamani: Since we aren't sure what could be the error here, status could be `500` and maybe a generic error saying, `there was an error, please try again in a moment` or something like that.

server/controllers/post/deleteById.js
tejas-hosamani: Yeah, but you are thinking in terms of UI. This is an API, anybody can send requests to API, with/without UI. So, opening the post is only an ideal case.

server/controllers/post/create.js
mittalyashu: Consider this possibility, if someone passes some additional data in the body request, then what.

server/controllers/post/create.js
mittalyashu: That's good, but, it's still not enough.

* Remove special characters
* Down case all characters
* Replace spaces with `-`

* Append random ID _(it's obvious so no need to add in the list)_

server/controllers/post/create.js
tejas-hosamani: It works as usual. That wouldn't be a problem at all.

server/controllers/post/deleteById.js
mittalyashu: Hmm... Let me test that SQL query locally.

server/routes/v1/posts.js
mittalyashu: > User should be logged in to view posts right?

All the posts are accessible to the member, regardless of the auth.

But they won't be able to comments on a post until they login.

server/controllers/post/updatePostById.js
mittalyashu: > Also, return error to user as well?

Yup, will look into that.

server/controllers/post/updatePostById.js
mittalyashu: > Check whether post exists before updating.

Hmm... Let me test that SQL query locally.

> Plus, data validation is necessary too.

Any thoughts on this one? @CodingCreate101

server/routes/v1/posts.js
tejas-hosamani: So, technically anybody on the internet can access this post as long as they have slug, correct?

mittalyashu: Right, after all it's a public site.

server/controllers/post/updatePostById.js
tejas-hosamani: Whatever is required for post creation should be required for update. Plus, all these:
```js
	const postId = req.params.postId;
	const postTitle = req.body.title;
	const bodyMarkdown = req.body.bodyMarkdown;
	const categoryId = req.body.categoryId;
	const statusId = req.body.statusId;
```
are required. Removed the slug. And I think we don't allow them to change slug? And maybe title too?

server/routes/v1/posts.js
tejas-hosamani: That's actually good. It makes things a little transparent.

server/controllers/post/updatePostById.js
mittalyashu: Yes, we only allow them to change title not the slug, since the slug will be auto-generated and based on the the url will also be updated.

server/controllers/post/create.js
mittalyashu: So, how should we do it 😅

server/controllers/post/create.js
mittalyashu: > It works as usual.

But, still there is slightest possibility. Right?

tejas-hosamani: No, I don't think so. Could you give a specific example?

server/controllers/post/create.js
tejas-hosamani: Okay.
```js
const slug = req.body.title.replace(/[^\w\s]/gi, '').toLowerCase().replace(' ', '-') + randomId
```

Try something like that.

mittalyashu: The problem with this regex query is that it will not recognise any double or triple spaces.

For example:

```
k          k
```

will be converted to `k-------k`

server/controllers/post/deleteById.js
mittalyashu:
```sql
SELECT 
    column_1 
FROM 
    table_1
    WHERE 
    EXISTS( SELECT 
                1 
            FROM 
                table_2 
            WHERE 
                column_2 = table_1.column_1);
```

We can do something like this to check the existence of the post, the query itself is very long.

mittalyashu: You're correct we need to find validate if the post exist or not but I think we can skip it for now.

What do you say? @CodingCreate101

server/controllers/post/updatePostById.js
mittalyashu: Does that makes sense? @CodingCreate101

server/controllers/post/updatePostById.js
mittalyashu: BTW, what kind of error type should we return to client side. @CodingCreate101

mittalyashu:
@CodingCreate101 Can you review the recent changes I have made?

PR #17, opened 2020-02-05
Post component

mittalyashu:
Showing number of upvotes, title, bodyMarkdown.

| design |
| -- |
| Screenshot_2020-02-05_at_25644_PM |

### DIFF
--- a/src/components/layout/Post.vue
+++ b/src/components/layout/Post.vue
@@ -1,0 +1,44 @@
+<template>
+	<div class="post">
+		<div class="post__voters">
+			<arrow-icon
+				class="post__voters-arrow"
+				:class="{ 'post__voters-vote': vote }"
+			/>
+			{{ voters }}
+		</div>
+		<div class="post__content">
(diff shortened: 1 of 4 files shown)


src/components/layout/Post.vue
```diff
@@ -0,0 +1,44 @@
...
+		<div class="post__voters">
+			<arrow-icon
+				class="post__voters-arrow"
+				:class="{ 'post__voters-vote': vote }"
+			/>
+			{{ voters }}
+		</div>
+		<div class="post__content">
+			<h3 class="post__content-title">{{ title }}</h3>
+			<p class="post__content-bodyMarkdown" v-html="bodyMarkdown" />
```
tejas-hosamani, line 12: Try converting markdown to HTML and then use something like `htmlData.replace(/(<([^>]+)>)/ig, "");` to remove all tags. IK you wanna do it later. Just leaving a suggestion.

mittalyashu: Technically this will be some kind of short description of around 20 to 30 words.

PR #162, opened 2020-10-12
Resolve "no-mutate-props"

HunterJS-bit:
Fix esling Resolve "no-mutate-props"  in Post.vue component
Other way to solve this would be to emmit data to parent component and do there mutation

### DIFF
--- a/frontend/src/components/post/Post.vue
+++ b/frontend/src/components/post/Post.vue
@@ -1,36 +1,26 @@
 <template>
-	<div class="post">
-		<Vote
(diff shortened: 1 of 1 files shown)


mittalyashu:
Thank you for your contribution @HunterJS-bit 

It looks like your changes is re-writing the whole file, can you only force-push the changes you have made.

HunterJS-bit:
@mittalyashu yeah sure

HunterJS-bit:
@mittalyashu Updated code

Issue #135, opened 2020-10-03
Resolve "no-mutate-props" (eslint error) for Post.vue component


mittalyashu:
The error is shown by vue eslint for "no-mutation-props", as we're modifying the props `post` when there an event call from `updateVoters`.

```js
updateVoters(voters) {
  // eslint-disable-next-line vue/no-mutating-props
  this.post.voters = voters;
}
```
[source](https://github.com/logchimp/logchimp/blob/master/frontend/src/components/post/Post.vue#L72-L75)

HunterJS-bit:
@mittalyashu Hi could I work on this issue ?

mittalyashu:
Hello @HunterJS-bit,

Sure, you can work on this issue.

HunterJS-bit:
@mittalyashu created pull request #162

HunterJS-bit:
@mittalyashu updated and created new pull request #163

HunterJS-bit:
@mittalyashu   can you please let me know if this PR is that you wanted.

mittalyashu:
@HunterJS-bit Apologies for delay and thank you for submitting the PR.

I will review it by today.

mittalyashu:
This issue is already resolved.

_Don't remember the PR, though 😅_

PR #163, opened 2020-10-13
fix: Resolve no-mutate-props (eslint error)

HunterJS-bit:
Resolved eslint error with mutating props #135

### DIFF
--- a/frontend/src/components/post/Post.vue
+++ b/frontend/src/components/post/Post.vue
@@ -52,19 +52,19 @@ export default {
 	},
 	data() {
 		return {
-			postData: this.post,
-		}
+			postData: this.post
+		};
 	},
 	components: {
 		Vote
 	},
 	watch: {
-    	post: {
-        	handler(newValue, oldValue) {
-        		this.postData = newValue;
-        	}
-      	}
-    },
+		post: {
+			handler(newValue, oldValue) {
+				this.postData = newValue;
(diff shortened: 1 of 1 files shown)


frontend/src/components/post/Post.vue
```diff
@@ -50,9 +50,21 @@ export default {
...
 	components: {
 		Vote
 	},
+	watch: {
+    	post: {
+        	handler(newValue, oldValue) {
+        		this.postData = newValue;
+        	}
+      	}
+    },
```
mittalyashu: Where are you using this post watch function?

mittalyashu: @HunterJS-bit

HunterJS-bit: @mittalyashu Because when I created `postData` will not be reactive when prop changes, it will not assign new value from `post` prop when it changes

frontend/src/components/post/Post.vue
```diff
@@ -71,7 +83,7 @@ export default {
 	methods: {
 		updateVoters(voters) {
 			// eslint-disable-next-line vue/no-mutating-props
```
mittalyashu:
```suggestion
```

Can you remove this line and see if eslint throws an error or not?

mittalyashu (approved):

PR #189, opened 2020-10-23
Reset password

mittalyashu:
**Designs**

| | screenshot |
| --- | :---: |
| Login | CleanShot 2020-10-23 at 10 45 21@2x |
| Forget password | CleanShot 2020-10-23 at 10 47 35@2x |
| Set new password | CleanShot 2020-10-23 at 10 48 06@2x |

Please share your feedback on the designs for password reset form.


server/controllers/auth/password/index.js
```diff
@@ -0,0 +1,8 @@
+const { readSync } = require("fs-extra");
```
mittalyashu:
```suggestion
```

Remove this unused code.

PR #556, opened 2021-10-06
Setup configs using env vars

mariotaddeucci:
Enabling configs using env vars. This method overwrite the configs file data.
This allows users to setup the development env and production easily, just changing the env vars without changing files or code.


mittalyashu:
Thank you @mariotaddeucci for your contribution.

There is already `--env` flag in the CLI, you can use to generate LogChimp configuration using environment variables https://logchimp.codecarrot.net/docs/cli/config/generate

PR #955, opened 2025-07-11
Fix/gh codespaces configuration

mittalyashu:


mittalyashu:
The codespaces running to run the server, postgres (db), and mail service.

But it still does not seems to automatically run the theme dev server after the codespaces is up and running.

Issue #967, opened 2025-07-20
Roadmaps loads/display twice


mittalyashu:
**Expected behaviour**

The roadmap show load only once.

**Screenshots**

![Image](https://github.com/user-attachments/assets/b5dddcbb-eb22-4765-8246-8261aa222745)

PR #968, opened 2025-07-22
Fix: Prevent duplicate roadmaps from loading on initial fetch

kris70lesgo:
This PR resolves an issue where the list of roadmaps was loading twice during the initial render, causing duplicated columns in the UI.

Changes made:

- Updated getRoadmaps method to conditionally push new data only on subsequent pages.
- Ensured roadmaps.value is replaced (not appended) on the first page load.

 Bug root cause:
The original logic always used roadmaps.value.push(...) without checking if it was the first page. This resulted in the same items being rendered multiple times on initial load.

 Result:
Now the roadmap list loads only once during the initial render and appends properly on scroll.
Related issue : [https://github.com/logchimp/logchimp/issues/967](url)

### DIFF
--- a/packages/theme/src/pages/Roadmaps.vue
+++ b/packages/theme/src/pages/Roadmaps.vue
@@ -46,15 +46,21 @@ async function getRoadmaps() {
 
   try {
     const response = await getAllRoadmaps();
-    roadmaps.value = response.data.roadmaps;
 
-    if (response.data.roadmaps.length) {
-			roadmaps.value.push(...response.data.roadmaps);
-			page.value += 1;
-			state.value = "LOADED"
-		} else {
-			state.value = "COMPLETED";
-		}
+    const newRoadmaps = response.data.roadmaps;
+
+    if (newRoadmaps.length > 0) {
+      if (page.value === 1) {
+        roadmaps.value = newRoadmaps;
+      } else {
+        roadmaps.value.push(...newRoadmaps);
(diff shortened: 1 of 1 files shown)


mittalyashu:
@kris70lesgo Whenever you working on a bug fix or a new feature it always best to create a new branch. 🙂

mittalyashu:
@kris70lesgo 

1. for writing simple git commits you don't have to add them in single/double quotes.
Here's one example where you might/can add quote: `feat(sitemap): add '/premium' URL to sitemap/pages`
You don't have to change this git commit message now, keep that in mind for future. 🙂

2. Can remove that merge commit, it think it doesn't need to part of commit history.

CleanShot 2025-07-23 at 16 28 34@2x

kris70lesgo:
thanks for the advice i will keep in mind and i have removed the merge commit

mittalyashu:
@kris70lesgo In the PR description you have added this:

> * Added type hint ref<unknown[]>([]) for clarity.
> * Improved scroll state management (LOADED, COMPLETED, ERROR).

But I don't see any related changes.

kris70lesgo:
Thanks for pointing that out! You're right — those changes were part of an earlier iteration but didn’t make it into the final commit. I’ll remove that from the PR description to keep things accurate. Appreciate the catch!

mittalyashu:
LGTM

PR #977, opened 2025-07-23
feat: bump tailwind v4

mittalyashu:
Start migrating from SASS to CSS modules or Tailwind


packages/theme/src/components/Header.vue
kris70lesgo, line 1: The use of Tailwind classes like bg-[--color-brand-color], w-full max-w-6xl mx-auto, flex items-center justify-between top-4 px-6 seems correct. However, ensure that the mt-2 class is intended for the nav div

packages/theme/src/components/ui/Avatar/AvatarStack.vue
kris70lesgo, line 1: The transition is mostly correct, but the border border-(--color-white) class might not work as expected if the --color-white variable is not defined in the CSS. Ensure that the color variables are correctly defined or replaced with Tailwind's color

mittalyashu: @kris70lesgo can you confirm on your end that you are able to see the white border in `AvatarStack.vue` file.

packages/theme/src/components/Header.vue
mittalyashu: That's a good observation.

The reason I didn't added `mt-2` class directly to `Navbar.vue` component because this css class does not internal change the styling/design of the component itself. It just adds a margin around the component itself, that's why I have added it outside of the component.

packages/theme/src/components/ui/Avatar/AvatarStack.vue
mittalyashu: @kris70lesgo 👆

kris70lesgo: sorry it took so long for me 
yeah i can see the white border in avatarstack.vue

kris70lesgo: i thinks its ready to merge

PR #1018, opened 2025-08-01
feat(theme/dashboard): improve sidebar HTML structure

mittalyashu:


mittalyashu:
@Vedant005 I am attaching a screen recording of an issue I am not able to pin-point the exact source of this issue.

Along the right-hand side of the sidebar there is some blue color glow.

https://github.com/user-attachments/assets/cc4572e4-8a5a-409d-a68c-a5d26685dcc4

Vedant005:
> @Vedant005 I am attaching a screen recording of an issue I am not able to pin-point the exact source of this issue.
> 
> Along the right-hand side of the sidebar there is some blue color glow.

I think this issue maybe be due to use of box-shadow or filter applied to any of its higher level container. I also see the blue color in `<body data-v-app="" style="--color-brand-color: #4B4D7C;">` but I checked and it does not lead to any use of box-shadow or filter

Well if you see in the video, when the sidebar is initially opened there is no glow effect and the effect appears after sometimes
If we are looking at any css effect then it could be the 
`<div class="overflow-hidden overflow-y-auto ..."> `

indicate that some browsers may show scrollbar glow effect if the scollbar is hidden or user has initialted touch/scroll.

I would suggest to try it with different browser as I think it's browser's visual feedback for the scrollable area, especially for an overlay scrollbar

mittalyashu:
@Vedant005 apart from this sidebar [bg color issue](https://github.com/logchimp/logchimp/pull/1018#issuecomment-3149948941).

Are we good to merge this PR?

Vedant005:
Yes we are good to merge this PR

mittalyashu:
@Vedant005 did you tested it on your machine?

Vedant005:
Yes I did test it. There was no bg color issue in mine

mittalyashu:
@Vedant005 Can you share the following where you've tested it:

- Browser version
- OS name and version
- Node version

mittalyashu:
| Browser | Screencast | Description |
| :--- |  --- | --- |
| Chrome | https://github.com/user-attachments/assets/49112579-62d3-4854-a06b-3fceca21238a | It seems to be happening because of `backdrop-blur-sm` css class in `packages/theme/src/components/ui/DialogBackdrop.vue` file. |
| Firefox | https://github.com/user-attachments/assets/347f4408-6707-4f1a-808a-6be3f36b7601 | Working as expected |
| Safari | https://github.com/user-attachments/assets/c7ff5792-2feb-49a4-8f09-e415dc664079 | Working as expected |

Why is it still happening? Still not sure.

PR #1020, opened 2025-08-03
feat(theme): display top public boards list on homepage

mittalyashu:


Vedant005:
The home page should also show "No boards available" if no public boards are created and I think the view all boards button should also be hidden in such case.

Rest changes are good

mittalyashu:
LGTM

PR #1021, opened 2025-08-03
feat: create Biome config for theme pkg

mittalyashu:
Need to confirm if running the Biome formatter and linter the CLI will automatically use the sub configuration file for respective directories?

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -20,8 +20,8 @@
     "includes": [
       "**",
       "!.idea",
-      "!packages/theme/dist",
-      "!packages/server/dist"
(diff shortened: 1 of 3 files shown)


mittalyashu:
This PR is closed by https://github.com/logchimp/logchimp/pull/1941, https://github.com/logchimp/logchimp/pull/1942, https://github.com/logchimp/logchimp/pull/1943

Biome as of September 17, 2026 does not fully support Vue.js.

To keep the Vue files code formatted and cleaned we have starts using Prettier.

PR #1028, opened 2025-08-04
fix: Add username length validation, length reference 20200906183548_…

systemmin:
[#990 ](https://github.com/logchimp/logchimp/issues/990)


systemmin:
recheck

PR #1057, opened 2025-08-15
fix: server crash when signing in with the same credentials

Ja4V8s28Ck:
This PR fixes the authentication logic to safely handle repeated sign in attempts with the same credentials without crashing the server.

**Changes made in**
- `packages/server/src/controllers/auth/setup.ts`
- `packages/server/src/controllers/auth/signup.ts`

### DIFF
--- a/packages/server/src/controllers/auth/setup.ts
+++ b/packages/server/src/controllers/auth/setup.ts
@@ -48,6 +48,10 @@ export async function setup(req, res, next) {
       name,
     });
 
+    // if user already exists, createUser returns null
+    if(!user)
+      return
+
     // set user as owner
     await database
       .update({
--- a/packages/server/src/controllers/auth/setup.ts
+++ b/packages/server/src/controllers/auth/setup.ts
@@ -49,8 +49,7 @@ export async function setup(req, res, next) {
     });
 
     // if user already exists, createUser returns null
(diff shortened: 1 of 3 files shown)


mittalyashu:
@Ja4V8s28Ck I was trying to add a test case previously in `packages/server/tests/integration/v1/auth/signup.spec.ts` but it's commented out.

No wonder this bug passed though.

While we're at it, can you add the test case for `USER_EXISTS` as well in this PR?

Ja4V8s28Ck:
> @Ja4V8s28Ck I was trying to add a test case previously in `packages/server/tests/integration/v1/auth/signup.spec.ts` but it's commented out.
> 
> No wonder this bug passed though.
> 
> While we're at it, can you add the test case for `USER_EXISTS` as well in this PR?

Sure, coming right up

mittalyashu:
@Ja4V8s28Ck Seems like you accidentally committed the up votes changes in this PR.

Ja4V8s28Ck:
> @Ja4V8s28Ck Seems like you accidentally committed the up votes changes in this PR.

Oh no

mittalyashu:
LGTM

mittalyashu:
Now the CI is working, yay!!!

PR #1058, opened 2025-08-15
test: added integration test for [DELETE] /api/board endpoint

### LINKED ISSUE #972: Write integration test for `[DELETE] /api/board` API endpoint
**Important**

* It should test the scenario where user with appropriate permissions should only be able to delete the board.

Ja4V8s28Ck:
Fixes #972 

Covers the [DELETE] /api/board endpoint with an integration test.

**Ensures that**
- Requests without valid `boardId` are rejected.
- Only users with proper permission(s) will be able to delete boards.

### DIFF
--- a/packages/server/tests/integration/v1/boards.spec.ts
+++ b/packages/server/tests/integration/v1/boards.spec.ts
@@ -155,3 +155,74 @@ describe("POST /api/v1/boards", () => {
   //     .returning(["userId"]);
   // });
 });
+
+// Delete boards by id
+describe("DELETE /api/v1/boards", () => {
+  it("should throw error 'NOT_ENOUGH_PERMISSION'", async () => {
+    const board = generateBoards();
+
+    await database.insert(board).into("boards");
+    const { user: authUser } = await createUser();
+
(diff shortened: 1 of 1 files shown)


packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -155,3 +155,74 @@ describe("POST /api/v1/boards", () => {
...
+      })
+      .into("permissions_roles");
+
+    await database
+      .insert({
+        id: uuid(),
+        role_id: newRoleId,
+        user_id: authUser.userId,
+      })
+      .into("roles_users");
```
mittalyashu: @Ja4V8s28Ck can you convert these DB operations into single DB trx?

Ref: https://github.com/logchimp/logchimp/pull/1059

Ja4V8s28Ck: Converted them to transactions

mittalyashu: Thanks

mittalyashu:
LGTM

PR #1074, opened 2025-08-18
chore(deps): bump faker v9.9.0

mittalyashu:


mittalyashu:
I don't see any other breaking change from Faker package that affects our codebase.

mittalyashu:
LGTM

PR #1077, opened 2025-08-18
fix: flaky integration tests

Ja4V8s28Ck:
Fixes the flakiness in the integration tests.


packages/server/tests/utils/db.ts
```diff
@@ -9,3 +11,22 @@ export const cleanDb = async () =>
...
+      username: "user_exists",
+    },
+    {
+      userId: uuidv4(),
+      email: "user_blocked@example.com",
+      password: hashPassword("strongPassword"),
+      username: "user_blocked",
+      isBlocked: true,
+    },
+  ]);
```
mittalyashu: Do we still need this piece of code?

Ja4V8s28Ck: Yes in the current test setup it is needed, did you remove `packages/server/src/database/seed/user.ts` ???

But we can remove it and run the test without seeding, that's also choice. If you want me to remove `seedDb` function, it's doable.

mittalyashu: I have already removed the file in [this PR](https://github.com/logchimp/logchimp/pull/1076/files#diff-c9a5c3d3abaee80bf1d32fef4d993e1d275a09178755cdf2cee1035aa11b2f92).

packages/server/tests/utils/db.ts
```diff
@@ -9,3 +11,22 @@ export const cleanDb = async () =>
     await trx.from("roadmaps").del();
     await trx.from("users").del();
   });
+
+export const seedDb = async () => {
+  await database.migrate.latest();
```
mittalyashu: What was the idea behind running the migration as a part of the test?

If you look into `.github/workflows/test.yml` file we're running the database migration separately.

packages/server/tests/utils/db.ts
Ja4V8s28Ck: Yeah, I noticed it after I wrote the tests. So, what is your choice? Do we keep the seeding file `packages/server/src/database/seed/user.ts` or remove it and find an alternative?

mittalyashu: The `packages/server/src/database/seed/user.ts` file has already been removed from `master` branch.

mittalyashu: The `packages/server/src/database/seed/user.ts` file has already been removed from `master` branch.

packages/server/tests/utils/db.ts
Ja4V8s28Ck: I must have missed to remove it after debugging.

Ja4V8s28Ck:
Integration test flakiness is temporarily fixed by changing some testcases. `file parallelism` and `--retry=1` flag in `vitest` are still an issue. `Can be reopened later, if this poses as a major issue`

PR #944, opened 2025-07-08
Add reusable domain blacklist check using environment variable

Shivanshee-sahu:
### Summary
This PR introduces a centralized utility function to validate blacklisted email domains using an environment variable. The function is reused across both `signup` and `login` controllers to ensure consistency and reduce code duplication.

---

### Changes Made
- ✅ Created `isDomainBlacklisted` function in `services/auth/domainBlacklist.js`
- ✅ Fetched `BLACKLISTED_DOMAINS` from `.env` and parsed it into a normalized array
- ✅ Used the function in:
  - `signup.js`: blocks signups from blacklisted domains
  - `login.js`: blocks logins from blacklisted domains
- ✅ Ensured domain matching is case-insensitive and trimmed

---

### Environment Variable Required
```env
BLACKLISTED_DOMAINS=example.com,example.org

PR #1078, opened 2025-08-18
Blacklist domain check using environment variable

Vedant005:
### What does this PR do?

- Introduces a centralized utility function to validate blacklisted email domains using an environment variable.
- Can be resued across apis where domain checks are needed.

### Changes made

- isDomainBlacklisted function in `packages/server/src/utils`
- used across singup, login, email : validate, verify, password: set, reset, setup, 
- Added EMAIL_DOMAIN_BLACKLISTED test case in login, singup integration tests
- Created unit test for domain blacklist logic

Closes #944 

## Summary by CodeRabbit

* **New Features**
  * Administrators can configure a domain blacklist to block signups/logins; attempts show a clear "Email domain is blacklisted" error.

* **Documentation**
  * Environment/config examples updated to include the domain blacklist setting and usage notes.

* **Tests**
  * Added unit and integration tests covering blacklist parsing, validation, and rejected-auth scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/theme/src/pages/Join.vue
+++ b/packages/theme/src/pages/Join.vue
@@ -144,6 +144,11 @@ async function join() {
       emailError.show = true;
       emailError.message = "Exists";
     }
+
+    if (error.response.data.code === "EMAIL_DOMAIN_BLACKLISTED") {
+      emailError.show = true;
+      emailError.message = "Email domain is blacklisted";
+    }
   } finally {
     buttonLoading.value = false;
   }
--- a/packages/theme/src/pages/Login.vue
+++ b/packages/theme/src/pages/Login.vue
@@ -143,7 +143,7 @@ async function login() {
 
     if (error.response.data.code === "EMAIL_DOMAIN_BLACKLISTED") {
       emailError.show = true;
-      emailError.message = "Email domain blacklisted";
+      emailError.message = "Email domain is blacklisted";
     }
 
     if (error.response.data.code === "INCORRECT_PASSWORD") {


Ja4V8s28Ck:
@Vedant005 Not sure, if I'm right. Rather than changing every files and adding `isDomainBlacklisted` function in all `auth controllers` (`signin.ts`, `login.ts`, `setup.ts` .. so on) , you could simply add a new middleware and just add it to the routes right? Correct me if I'm wrong @mittalyashu.

Vedant005:
> add a new middleware and just add it to the routes 

@Ja4V8s28Ck  Yes I thought of that but I think the middleware would just do a simple call to this util function and then we can use it in the routes. was also thinking on how much it would affect the efficiency of apis as some of them already have 2-3 middlewares.

Well the previous author kind of used it as a util function and @mittalyashu approved of that I think.

mittalyashu:
> @Vedant005 Not sure, if I'm right. Rather than changing every files and adding `isDomainBlacklisted` function in all `auth controllers` (`signin.ts`, `login.ts`, `setup.ts` .. so on) , you could simply add a new middleware and just add it to the routes right? Correct me if I'm wrong @mittalyashu.

Sure, give it a try. If moving the logic to a middleware simplifies the code.
Because the order in which data validation is done in a controller is important as well.

mittalyashu:
> > add a new middleware and just add it to the routes
> 
> @Ja4V8s28Ck Yes I thought of that but I think the middleware would just do a simple call to this util function and then we can use it in the routes. was also thinking on how much it would affect the efficiency of apis as some of them already have 2-3 middlewares.

Adding a middleware should not impact the performance in Node.js to great extend.

> Well the previous author kind of used it as a util function and @mittalyashu approved of that I think.

The previous author treated it more like a utility function, which is also a valid approach.

In general, there are multiple ways to structure this — from simple util calls to dedicated middlewares. It really depends on what keeps the codebase more maintainable and consistent. **That’s part of the learning process.**

Ja4V8s28Ck:
Can you check `Test / API / Unit Tests`, there is an import error.

mittalyashu:
@Vedant005 For now you can ignore these "Missing rate limiting" messages.

packages/server/src/errorResponse.json
```diff
@@ -21,6 +21,9 @@
   "api": {
     "authentication": {
       "invalidEmail": "Enter valid email",
+      "invalidEmailFormat": "Invalid email format",
+      "invalidEmailDomain": "Invalid email format",
```
mittalyashu, line 33: @Vedant005 both of the keys has same value.

packages/server/tests/integration/v1/auth/login.spec.ts
```diff
@@ -10,6 +10,19 @@ import database from "../../../../src/database";
 import { hashPassword } from "../../../../src/utils/password";
 
 describe("POST /api/v1/auth/login", () => {
+  it("should throw EMAIL_DOMAIN_BLACKLISTED", async () => {
```
mittalyashu, line 10:
```suggestion
  it('should throw error "EMAIL_DOMAIN_BLACKLISTED"', async () => {
```

packages/server/tests/integration/v1/auth/signup.spec.ts
```diff
@@ -6,6 +6,19 @@ import app from "../../../../src/app";
 import database from "../../../../src/database";
 
 describe("POST /api/v1/auth/signup", () => {
+  it("should throw EMAIL_DOMAIN_BLACKLISTED", async () => {
```
mittalyashu, line 9:
```suggestion
  it('should throw error "EMAIL_DOMAIN_BLACKLISTED"', async () => {
```

mittalyashu (changes requested):

mittalyashu:
@Vedant005 If the file does not include the changes in scope of this PR, you can ignore the changes like `// @ts-expect-error`.

We can handle that all-together in a separate PR.

packages/server/src/middlewares/domainBlacklist.ts
```diff
@@ -0,0 +1,80 @@
...
+        return;
+      }
+      result.add(domain);
+    });
+
+  return result;
+};
+
+const getBlacklistedDomains = () =>
+  parseBlacklistedDomains(process.env.LOGCHIMP_BLACKLISTED_DOMAINS || "");
```
mittalyashu: @Vedant005 can you check that this variable only gets computed once per server reboot.

If we try to re-compute and process this function call for every single API request that would not be performant _(abet marginal difference)_

packages/server/src/app.ts
```diff
@@ -1,11 +1,14 @@
...
 import cors from "cors";
+import dotenv from "dotenv";
 
 import routes from "./routes";
 
 const app = express();
 app.disable("x-powered-by");
 
+dotenv.config();
+
```
mittalyashu: Strange we were not importing `dotenv` module.

I think now that we were prominently using Docker even for local development, we didn't noticed the need of importing it.

Even though we had `dotenv` package installed.

packages/server/tests/integration/v1/auth/login.spec.ts
```diff
@@ -10,6 +10,19 @@ import database from "../../../../src/database";
...
+    process.env.LOGCHIMP_BLACKLISTED_DOMAINS =
+      "example.com, test.com, spam.com, badsite.org";
```
mittalyashu, line 12: Can you check if there is a better to set the environment variable.

for example `process.setEnv()`, this is just my assumption, might be wrong.

packages/theme/src/pages/setup/CreateAccount.vue
```diff
@@ -186,7 +186,10 @@ async function createAccount() {
 			serverError.value = true;
 		}
 
-		console.error(error);
```
mittalyashu: @Vedant005 you have removed this error logging statement.

packages/server/src/app.ts
Vedant005: Yeah I thought why was the case.

packages/theme/src/pages/setup/CreateAccount.vue
Vedant005: Is it necessary? I thought we don't need it.

packages/server/tests/integration/v1/auth/login.spec.ts
Vedant005: I checked, there is no better alternative to process.env

packages/server/src/middlewares/domainBlacklist.ts
Vedant005: What if we used `Set` so that unique values are stored and use getBlacklistedDomains as a variable

```
const getBlacklistedDomains : Set<string> = parseBlacklistedDomains(
  process.env.LOGCHIMP_BLACKLISTED_DOMAINS || ""
);
```

packages/server/src/middlewares/domainBlacklist.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+  return result;
+};
+
+export class BlacklistManager {
+  private cachedBlacklist: Set<string> | null = null;
+
+  constructor(
+    private rawDomains: string = process.env.LOGCHIMP_BLACKLISTED_DOMAINS || "",
```
mittalyashu: Move the assignment of the property `cachedBlacklist` inside the constructor.

packages/server/src/middlewares/domainBlacklist.ts
```diff
@@ -0,0 +1,98 @@
...
+  ) {}
+
+  getBlacklistedDomains() {
+    if (!this.cachedBlacklist) {
+      this.cachedBlacklist = parseBlacklistedDomains(this.rawDomains);
+    }
+    return this.cachedBlacklist;
```
mittalyashu: Can you reverse this check?

If there is a cache, return the cache, else compute the fresh-new value.

packages/theme/src/pages/setup/CreateAccount.vue
mittalyashu: Yeah, as of now we have not decided on one "official" way to adding support for logging.

So, please revert back this change.

packages/server/tests/integration/v1/auth/login.spec.ts
mittalyashu: Ok, thanks for the information.

packages/server/src/app.ts
mittalyashu: For now, can you please revert back this change. I'll look into this separately.

packages/server/src/errorResponse.json
mittalyashu: @Vedant005 Once you've updated the code for related review, it's best practice to share an quick update in the comment or mark the review as "Resolve conversation". 🙂

Vedant005: Understood

packages/server/src/middlewares/domainBlacklist.ts
Vedant005: Done

packages/server/tests/unit/domainBlacklist.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    expect(isValidDomain(null)).toBeFalsy();
+    //@ts-ignore
+    expect(isValidDomain(undefined)).toBeFalsy();
+    //@ts-ignore
+    expect(isValidDomain(123)).toBeFalsy();
+    //@ts-ignore
+    expect(isValidDomain({})).toBeFalsy();
+    //@ts-ignore
+    expect(isValidDomain([])).toBeFalsy();
+  });
```
mittalyashu, line 115: @dinirichard Should we add this test case or not?

packages/server/src/errorResponse.json
mittalyashu: @Vedant005 Both the key have the same value. Is there any specific reason?

mittalyashu:
Merging the PR as-is, to an internal branch to run the CI pipeline. The pipeline is failing in this PR because of the froked-repository.

mittalyashu:
Merging the PR as-is, to an internal branch to run the CI pipeline. The pipeline is failing in this PR because of the froked-repository.

PR #1087, opened 2025-08-20
run mailhog service in test workflow

mittalyashu:


mittalyashu:
The integration tests failing in this PR is due to flakyness.

PR #1093, opened 2025-08-21
Changes 

Snigdha977:
Suggested some changes on features of the existing log chimp software


Snigdha977:
I reviewed the guide and noticed a few areas for improvement. The Get Started button currently leads to the guidelines rather than a working prototype, which can be confusing for new users. In the Getting Started section, only the Robust REST APIs feature is explained in detail, so adding brief notes for the other features would create better balance. Including key CLI commands directly in the guide could also help developers get started faster. Finally, making the principles and roadmap more interactive with visuals would improve clarity and engagement.

PR #1101, opened 2025-08-23
fix: docker compose

Ja4V8s28Ck:
Fixes docker compose dev build for contribution. ~Bumped up the postgres image from `12` to `12.4`~

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -35,7 +35,12 @@ services:
       args:
         DOCKER_BUILDKIT: 1
     volumes:
-      - ./:/app
+      - .:/app
+      - /app/node_modules
+      - /app/packages/server/node_modules
+      - /app/packages/types/node_modules
+      - /app/packages/server/dist
+      - /app/packages/types/dist
     depends_on:
       - db
     networks:
--- a/docker/local/server.Dockerfile
+++ b/docker/local/server.Dockerfile
@@ -15,15 +15,12 @@ RUN go install github.com/air-verse/air@latest;
 WORKDIR /app
 COPY ./packages/types/ ./packages/types/
 COPY ./packages/server/ ./packages/server/
-COPY ./package.json ./pnpm-lock.yaml ./
-
-WORKDIR /app/packages/server
+COPY ./package.json ./pnpm-lock.yaml ./pnpm-workspace.yaml ./
 
 ENV NODE_ENV=development
 
-RUN \
-    npm i -g pnpm; \
-    pnpm install; \
-    pnpm --filter="@logchimp/types" build;
+RUN npm i -g pnpm && pnpm i && pnpm --filter="@logchimp/types" build
+
+WORKDIR /app/packages/server
 
 CMD ["air", "-c", ".air.toml"]
--- a/docker/local/server.Dockerfile
+++ b/docker/local/server.Dockerfile
@@ -19,7 +19,9 @@ COPY ./package.json ./pnpm-lock.yaml ./pnpm-workspace.yaml ./
 
 ENV NODE_ENV=development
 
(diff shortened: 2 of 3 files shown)


docker-compose.dev.yml
```diff
@@ -34,8 +34,6 @@ services:
       context: .
       args:
         DOCKER_BUILDKIT: 1
-    volumes:
-      - ./:/app
```
mittalyashu: @Ja4V8s28Ck any specific reason why removed this?

docker-compose.dev.yml
```diff
@@ -2,7 +2,7 @@ name: "LogChimp-Dev"
 
 services:
   db:
-    image: postgres:12
+    image: postgres:12.4
```
mittalyashu: @Ja4V8s28Ck Can you please revert back this change, we will bump to minor version in a separate PR.

docker/local/server.Dockerfile
```diff
@@ -15,15 +15,14 @@ RUN go install github.com/air-verse/air@latest;
...
 
 ENV NODE_ENV=development
 
-RUN \
-    npm i -g pnpm; \
-    pnpm install; \
-    pnpm --filter="@logchimp/types" build;
+RUN npm i -g pnpm
+RUN pnpm i
+RUN pnpm --filter="@logchimp/types" build
```
mittalyashu: You can still merge that into a single command.

docker-compose.dev.yml
Ja4V8s28Ck: The reason is that, this command is overwriting the container which is one of the reason why we got an error `ts-node` was not found.

docker/local/server.Dockerfile
Ja4V8s28Ck: Sure, 😅 writing separately looked cleaner, that's why.

I reverted it back to the earlier state, its causing some irregularity. I have nested all the functions together with `&&`

docker-compose.dev.yml
Ja4V8s28Ck: Sure, we can bump up the version another day

Ja4V8s28Ck:
@mittalyashu Before you merge, test it in a new clone of `logchimp`.

docker-compose.dev.yml
mittalyashu: The volume mount is important to use Docker for local development.
This allows the edits from the host machine goes to the container which in-turns reload the dev server.

Removing the volume amount will make the Docker container static, _just like running in prod_.

Ja4V8s28Ck: Oh, but if we are overwriting the entire `app` folder in the local system with that of docker build, it wont be doing any good.

Something that I can think right now is that, we can manually mount the files and folders that are important for contribution.

mittalyashu: > but if we are overwriting the entire app folder in the local system with that of docker build, it wont be doing any good.

Durning the docker build, it's always the host directory mount inside Docker container.

Basically it's a 2-way read:write.

At Docker runtime, the container can make overwrite whatever is on the host (on the mapped-path)

Ja4V8s28Ck: Yes you are right, but here is what I think and this is what I observer, right after the db and mailhog is done building, first the dockerbuild runs for the server where all the dependencies are being installed and setup. Right after that the volume is mounted from the app folder in host system which overlaps all the changes.

mittalyashu:
The PR LGTM

@Ja4V8s28Ck a review from your end before we merge it.

Ja4V8s28Ck:
> The PR LGTM
> 
> @Ja4V8s28Ck a review from your end before we merge it.

@mittalyashu Everything works as intended. LGTM

PR #1104, opened 2025-08-24
feat: improve TS types

mittalyashu:


packages/server/src/express.d.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface ExpressRequestContext<
+  P = core.ParamsDictionary,
+  ResBody = any,
+  ReqBody = any,
+  ReqQuery = core.Query,
+  Locals extends Record<string, any> = Record<string, any>,
+> extends Request<P, ResBody, ReqBody, ReqQuery, Locals> {
+  ctx: {
+    roadmap?: IRoadmapPrivate;
+  };
```
mittalyashu: Instead of directly passing the data to `req` object.
I've created a `ctx` object to pass data more easily due to TS types.

I tried declaring the types in `Express.Request` using global namespace but for some reason that didn't worked.

```ts
declare global {
  namespace Express {
    interface Request {
      roadmap?: IRoadmapPrivate;
    }
  }
}
```

PR #1105, opened 2025-08-25
fix: roadmaps load lazy

### LINKED ISSUE #1071: Lazy load roadmaps only on horizontal scroll
In the current implementation the roadmaps API does support cursor pagination but the public roadmaps page loads all the roadmap at once.

As the public roadmap page has to subsequently fetch all the posts per roadmap, it can signifiently put load on the backend and database.

To avoid that, we need to implement lazy load on public roadmap page.

We can limit the cursor pagination to 4 or 5 items per request for public roadmap page.

Ja4V8s28Ck:
Fixes #1071

**Changes Made**
- Current cursor pagination is set to 3. AKA only 3 roadmaps are shown after landing on the page.
- Roadmaps are lazy loaded based on scrolling.

### DIFF
--- a/packages/theme/src/pages/Roadmaps.vue
+++ b/packages/theme/src/pages/Roadmaps.vue
@@ -1,10 +1,11 @@
 <template>
   <!-- Show roadmaps grid only when we have roadmaps -->
   <div
-    v-if="roadmaps.length > 0" ref="roadmapElement"
+    v-if="roadmaps.length > 0"
+    ref="roadmapElement"
     :class="[
       'overflow-x-auto h-[500px]',
-      'grid grid-flow-col gap-x-4 md:gap-x-6 auto-cols-[minmax(22rem,24rem)]'
+      'grid grid-flow-col gap-x-4 md:gap-x-6 auto-cols-[minmax(22rem,24rem)]',
     ]"
   >
     <roadmap-column
@@ -31,24 +32,27 @@ import RoadmapColumn from "../ee/components/roadmap/RoadmapColumn.vue";
 const { get: siteSettings } = useSettingStore();
 
 // Cursor-based pagination state
+const pageSize = 4;
 const roadmapElement = useTemplateRef<HTMLElement>("roadmapElement");
 const roadmaps = ref<IRoadmap[]>([]);
-let roadmapList: IRoadmap[];
-const roadmapIndex = ref<number>(0);
+const endCursor = ref<string | undefined>();
+const hasNextPage = ref<boolean>(false);
 
-async function getRoadmaps() {
+async function getRoadmaps(after: string | undefined) {
   try {
-    const response = await getAllRoadmaps();
+    const response = await getAllRoadmaps({
+      first: pageSize.toString(),
+      after: after == null ? undefined : after,
+    });
 
     const paginatedData: IPaginatedRoadmapsResponse = response.data;
-    roadmapList = paginatedData.results;
+    const roadmapList = paginatedData.results;
 
     if (roadmapList.length > 0) {
-      // Initializing the only 3 roadmaps
-      for (; roadmapIndex.value < 3; ) {
-        roadmaps.value.push(roadmapList[roadmapIndex.value++]);
-      }
+      roadmaps.value.push(...roadmapList);
     }
+    endCursor.value = paginatedData.page_info.end_cursor || undefined;
+    hasNextPage.value = paginatedData.page_info.has_next_page;
   } catch (err) {
     console.error("Error fetching roadmaps:", err);
   }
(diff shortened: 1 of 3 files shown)


mittalyashu:
@Ja4V8s28Ck You can use [VueUse InfiniteScroll](https://vueuse.org/core/useInfiniteScroll/#useinfinitescroll) with horizontal scrolling.

packages/theme/src/pages/Roadmaps.vue
```diff
@@ -104,4 +117,22 @@ useHead({
...
   name: "Roadmaps",
 });
+
+onUpdated(() => {
+  observer = new IntersectionObserver(
+    (entries) => {
+      if (entries[0].isIntersecting) {
+        if (roadmapCount.value < roadmapArr.value.length) {
+          roadmaps.value.push(roadmapArr.value[roadmapCount.value++]);
+          console.log("@31");
```
mittalyashu:
```suggestion
```

mittalyashu:
The infinite scroll partially works.

For instance, if I set the following code to fetch roadmaps with 4 items in each page then it does not make the next roadmap API call.
```
const response = await getAllRoadmaps({
      first: 4
    });
```

Ja4V8s28Ck:
> The infinite scroll partially works.
> 
> For instance, if I set the following code to fetch roadmaps with 4 items in each page then it does not make the next roadmap API call.
> 
> ```
> const response = await getAllRoadmaps({
>       first: 4
>     });
> ```

I tested this case and when I scrolled I got an error 🤔 Did you get an error?
L65 is causing that issue `return roadmapIndex.value <= roadmapList.length;`, its supposed to be `return roadmapIndex.value < roadmapList.length;` _zero_based_indexing_

mittalyashu:
Yes, I made that change.

> canLoadMore: () => {
>      return roadmapIndex.value <= roadmapList.length;
>  },

Even before making the above change it was still throwing the error. _(sharing the error screenshot below)_

| Screenshot |
| --- |
| CleanShot 2025-09-03 at 18 53 55@2x |

Ja4V8s28Ck:
> Yes, I made that change.
> 
> > canLoadMore: () => {
> > return roadmapIndex.value <= roadmapList.length;
> > },
> 
> Even before making the above change it was still throwing the error. _(sharing the error screenshot below)_
> 
> Screenshot
> CleanShot 2025-09-03 at 18 53 55@2x

This is the same error I got, but I fixed it by the changing L65. Yes even before you formatted the code, it still gave an error. I made an error.

https://github.com/user-attachments/assets/cdd95147-3412-4865-af60-e9966ba8e925

mittalyashu:
For that you have to use the cursor pagination to fetch the next batch of the roadmaps.

Ja4V8s28Ck:
> For that you've have to use the cursor pagination to fetch the next batch of the roadmaps.

At the moment, `getAllRoadmaps()` fetches all the roadmaps in single request and then the posts associated with the roadmaps are lazy loaded. If we use `getAllRoadmaps`, won't it be sending multiple requests?? (considering the number of roadmaps is < 100). But if the roadmaps are more than 100, we should be fixing the UI before we use cursor pagination right?? or atleast limit the number of roadmaps shown in the public view, Horizontal scrolling wont be good if the number of roadmaps are any more than 50, we have first experience with `validator.js` on horizontal scrolling 😅

mittalyashu:
> > For that you've have to use the cursor pagination to fetch the next batch of the roadmaps.
> 
> At the moment, `getAllRoadmaps()` fetches all the roadmaps in single request and then the posts associated with the roadmaps are lazy loaded. If we use `getAllRoadmaps`, won't it be sending multiple requests?? (considering the number of roadmaps is < 100). But if the roadmaps are more than 100, we should be fixing the UI before we use cursor pagination right?? or atleast limit the number of roadmaps shown in the public view, Horizontal scrolling wont be good if the number of roadmaps are any more than 50, we have first experience with `validator.js` on horizontal scrolling 😅

You're right about that and for that we've cursor pagination.

In the example of roadmaps < 100, we will fetch roadmaps in batches by setting the page size to 4 or 5.

```
const response = await getAllRoadmaps({
   first: 4
});
```

and once the horizontal view is scrolled 90% (technically its called [threshold value](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#threshold)) using infinite scroll; the next 4 roadmaps will be fetched.

If you look at the API response of the `/api/v1/roadmaps`, it will look something like this:

```ts
export interface PageInfo {
  count: number;
  current_page: number;
  has_next_page: boolean;
  end_cursor: string | null;
  start_cursor: string | null;
}

export interface CursorPaginatedResponse<T> {
  results: T[];
  page_info: PageInfo;
  total_pages?: number | null;
  total_count?: number | null;
}
```

Look into `PageInfo` type.

Ja4V8s28Ck:
> Ready for Review.

mittalyashu:
LGTM

PR #1111, opened 2025-08-26
feat(theme): display logchimp version

mittalyashu:


mittalyashu:
I have tested how the `docker-compose.yml` theme service environment variables will be injected into the application.

PR #1112, opened 2025-08-26
feat: New utility function to create role with permissions for tests

Vedant005:
### What does this PR do?

Introduces new utility function to create role with permissions to be used for testing.
This avoids writing duplicate code for every test case where this utility is needed


packages/server/tests/utils/createRoleWithPermissions.ts
```diff
@@ -0,0 +1,53 @@
...
+  roleId?: string,
+  userId?: string,
+): Promise<string> {
+  const finalRoleId = roleId ?? uuidv4();
+
+  await database.transaction(async (trs) => {
+    if (!roleId) {
+      await trs("roles").insert({
+        id: finalRoleId,
+        name: permissions.map((p) => `${p.type}:${p.action}`).join(", "),
```
mittalyashu: @Vedant005 Can you cross check the database column max varchar for the `name` column.

Vedant005: @mittalyashu It's 30

Vedant005: Well if take all the permissions then it totals to 413 chars, so we could increase the size to 500 or something

mittalyashu: It does not make sense to increase the database varchar just to fulfil a test case.

Instead we should update the test itself. 🙂

packages/server/tests/utils/createRoleWithPermissions.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type PermissionInput = {
+  type: string;
+  action: string;
+};
+
+export async function createRoleWithPermissions(
+  permissions: PermissionInput[],
+  roleId?: string,
+  userId?: string,
```
mittalyashu: When defining the function argument it's important to note which arguments are important and which are not, and accordingly we mark them undefine-able.

Here `userId` is important, then `permissions`.

for the 3rd argument can you change it to an `role` object with the following properties:
1. roleId?: string
2. roleName?: string

packages/server/tests/utils/createRoleWithPermissions.ts
Vedant005: Yes but what if in we want to create a role with permissions like : "board:read, board:update, board:create" which exceeds  30 limit. So how can we handle that?

packages/server/tests/utils/createRoleWithPermissions.ts
Vedant005: Made changes

mittalyashu:
@Vedant005  Before we merge the PR, can you do a quick check, if there is any other place where this function can be re-used.

mittalyashu:
@Vedant005 Are we good to merge this PR?

Vedant005:
> @Vedant005 Are we good to merge this PR?

Yes, we are good to merge.

packages/server/tests/integration/v1/roles.spec.ts
```diff
@@ -27,41 +26,11 @@ describe("GET /api/v1/roles", () => {
...
-          id: uuid(),
-          role_id: roleId,
-          user_id: user.userId,
-        })
-        .into("roles_users");
-    });
+    await createRoleWithPermissions(
+      user.userId,
+      [{ type: "role", action: "read" }],
+      { roleId: undefined, roleName: "CreateRole" },
```
mittalyashu:
```suggestion
      { roleName: "CreateRole" },
```

packages/server/tests/utils/createRoleWithPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+export async function createRoleWithPermissions(
+  userId: string,
+  permissions: PermissionInput[],
+  role?: RoleInput,
+): Promise<string> {
+  const roleId = role?.roleId ?? uuidv4();
+
+  const roleName =
+    role?.roleName ??
+    permissions.map((p) => `${p.type}:${p.action}`).join(", ");
```
mittalyashu: @Vedant005 As we've discussed previously.
This will cause a db error for longer names.

Can you handle this issue?

packages/server/tests/utils/createRoleWithPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+type RoleInput = {
+  roleId?: string;
+  roleName?: string;
+};
+
+export async function createRoleWithPermissions(
+  userId: string,
+  permissions: PermissionInput[],
```
mittalyashu: To make ease of use of this function, can you change the permissions args data type `TPermissions[]` from `@logchimp/types`

and handle the `type` and `action` permission search inside the function itself.

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -188,45 +152,11 @@ describe("DELETE /api/v1/boards", () => {
...
-          id: uuid(),
-          role_id: newRoleId,
-          user_id: authUser.userId,
-        })
-        .into("roles_users");
-    });
+    await createRoleWithPermissions(
+      authUser.userId,
+      [{ type: "board", action: "destroy" }],
+      { roleId: undefined, roleName: "Board destroyer" },
```
mittalyashu:
```suggestion
      { roleName: "Board destroyer" },
```

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -74,46 +73,11 @@ describe("GET /boards/search/:name", () => {
...
-          role_id: newRoleId,
-          user_id: authUser.userId,
-        })
-        .into("roles_users");
-    });
-
+    await createRoleWithPermissions(
+      authUser.userId,
+      [{ type: "board", action: "read" }],
+      { roleId: undefined, roleName: "Board Reader" },
```
mittalyashu:
```suggestion
      { roleName: "Board Reader" },
```

mittalyashu (changes requested):

packages/server/tests/utils/createRoleWithPermissions.ts
Vedant005: Should I truncate it? That would solve this issue

mittalyashu: Yes

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -188,45 +150,13 @@ describe("DELETE /api/v1/boards", () => {
...
-        .insert({
-          id: uuid(),
-          role_id: newRoleId,
-          user_id: authUser.userId,
-        })
-        .into("roles_users");
-    });
+    await createRoleWithPermissions(
+      authUser.userId,
+      ["board:destroy", "board:create", "board:update"],
```
mittalyashu: @Vedant005 Do we need `"board:create", "board:update"` permission for this test case?

Vedant005: Done

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -193,44 +155,8 @@ describe("DELETE /api/v1/boards", () => {
...
-      // assign the role to newly created user
-      await trx
-        .insert({
-          id: uuid(),
-          role_id: newRoleId,
-          user_id: authUser.userId,
-        })
-        .into("roles_users");
+    await createRoleWithPermissions(authUser.userId, ["board:destroy"], {
+      roleName: "Board destroyer",
```
mittalyashu, line 159: Hahah, that's a good name 🤣

gif

packages/server/tests/integration/v1/roles.spec.ts
```diff
@@ -27,40 +26,8 @@ describe("GET /api/v1/roles", () => {
...
-
-      await trx
-        .insert({
-          id: uuid(),
-          role_id: roleId,
-          user_id: user.userId,
-        })
-        .into("roles_users");
+    await createRoleWithPermissions(user.userId, ["role:read"], {
+      roleName: "CreateRole",
```
mittalyashu: We're assigning the `role:read` permission can you give an appropriate name accordingly.

mittalyashu (changes requested):
Feedback

packages/server/tests/integration/v1/roles.spec.ts
Vedant005: How about spectator or role reader?

mittalyashu:
LGTM

PR #1114, opened 2025-08-27
chore: add types-watcher service in docker-compose.dev.yml

mittalyashu:


mittalyashu:
For now, this might be bit overkill 😂 😬

PR #1121, opened 2025-08-29
Added test for Posts Slug (/posts/slug)

dinirichard:


mittalyashu:
@dinirichard For the business logic code you can continue to use `snakeCase` for variable and property names.

packages/server/tests/integration/v1/posts.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        userId: ""
+      }
+    );
+
+    expect(response.headers["content-type"]).toContain("application/json");
+    expect(response.status).toBe(200);
+    expect(response.body.post.slug).toEqual(post.slug);
+    expect(response.body.post.board.boardId).toEqual(board.boardId);
+    expect(response.body.post.roadmap.id).toEqual(roadmap.id);
+    expect(response.body.post.author.userId).toEqual(authUser.userId);
```
mittalyashu: @dinirichard You can create a common variable like:

```
const data = response.body.post
```

dinirichard:
My bad, I'll get it fixed.

packages/server/tests/utils/generators.ts
```diff
@@ -2,14 +2,19 @@ import { faker } from "@faker-js/faker";
 import { v4 as uuid } from "uuid";
 import generatePassword from "omgopass";
 import { nanoid } from "nanoid";
-
 import {
```
mittalyashu:
```suggestion
import { nanoid } from "nanoid";

import {
```

packages/server/tests/integration/v1/posts.spec.ts
mittalyashu: @dinirichard 👆

packages/server/tests/integration/v1/posts.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('should throw error "POST_NOT_FOUND"', async () => {
+    await cleanDb();
+
+    const response = await supertest(app)
+    .post("/api/v1/posts/slug")
+    .send(
+      {
+        slug: "dolores-ipsa-mKTAvagnq3xaZYaag2pU",
+        userId: "",
```
mittalyashu: @dinirichard Any specific reason you've explicitly added `userId` with empty string?

mittalyashu (changes requested):

packages/server/tests/integration/v1/posts.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { cleanDb } from "../../utils/db";
+
+describe("POST Slug /api/v1/posts/slug", () => {
+  afterEach(async () => {
+    await cleanDb();
+  });
+
+  it('should throw error "POST_NOT_FOUND"', async () => {
+    await cleanDb();
+
```
mittalyashu:
```suggestion
```

packages/server/tests/integration/v1/posts.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  board as generateBoards,
+  roadmap as generateRoadmap
+} from "../../utils/generators";
+import { createUser } from "../../utils/seed/user";
+import { cleanDb } from "../../utils/db";
+
+describe("POST Slug /api/v1/posts/slug", () => {
+  afterEach(async () => {
+    await cleanDb();
+  });
```
mittalyashu: The overall goal for tests is to have confidence that they can run in isolation without relying on manually seeded data (and use faker pkg) so that we don't have to clean the database.

packages/server/tests/integration/v1/posts.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      isVerified: true,
+    });
+
+    const post = generatePost({
+      userId: authUser.userId,
+      boardId: board.boardId,
+      roadmapId: roadmap.id,
+    });
+
+    await database.insert(post).into("posts");
```
mittalyashu: Can you create utility function for each (roadmaps, boards, posts) where it would fake the data and then insert it to the database as well.

Just like how we have `createUser` function, this way we can stop relying on functions like `generateRoadmap()`

dinirichard: Sure, but I'll still keep the generate functions.

packages/server/tests/integration/v1/posts.spec.ts
dinirichard: because, i noticed it can get the post without `userId` or authentication and i wanted the test to showcase that feature/flaw.

packages/server/tests/integration/v1/posts.spec.ts
mittalyashu: > but I'll still keep the generate functions.

Ok

packages/server/tests/integration/v1/posts.spec.ts
mittalyashu: > i noticed it can get the post without `userId` or authentication and i wanted the test to showcase that feature/flaw.

Okay 🙂

Can you add a short & simple comment above each `userId` property?

dinirichard (commented):
ready for review

Issue #1129, opened 2025-09-01
Unexpected double invocation of functions while editing posts.


Ja4V8s28Ck:
**Describe the bug**
`selectBoard` and `selectRoadmap` function in `packages/theme/src/pages/dashboard/posts/_slug/Index.vue` runs twice

**Screen recording**

https://github.com/user-attachments/assets/ee9b5f0d-53a3-490f-ace8-f9882a0e3595

PR #1136, opened 2025-09-01
fix: assign roadmap to post

Ja4V8s28Ck:
Patches the existing issue where the roadmaps can't be assigned to a post.

**Changes made**
- `packages/theme/src/pages/dashboard/posts/_slug/Index.vue`

Closes https://github.com/logchimp/logchimp/issues/1129

### DIFF
--- a/packages/theme/src/pages/dashboard/posts/_slug/Index.vue
+++ b/packages/theme/src/pages/dashboard/posts/_slug/Index.vue
@@ -295,7 +295,9 @@ async function suggestRoadmap(event: Event) {
 function selectBoard(index: number) {
   const item = boards.suggestions[index];
 
-  Object.assign(post.board, item);
+  Object.assign(post, {
+    board: item,
+  });
   boards.search = "";
   boards.suggestions = [];
 }
--- a/packages/theme/src/components/board/BoardSuggestion.vue
+++ b/packages/theme/src/components/board/BoardSuggestion.vue
@@ -19,6 +19,7 @@ interface Props {
 }
 
 defineProps<Props>();
+defineEmits(["click"]);
 </script>
 
 <style lang='sass'>


mittalyashu:
@Ja4V8s28Ck now that I think about it we also need to add support for un-assign roadmaps and boards.

---

I tested this issue as well and by explicitly defining a click emit event fixes this issue.

Ja4V8s28Ck:
> @Ja4V8s28Ck now that I think about it we also need to add support for un-assign roadmaps and boards.
> 
> I tested this issue as well and by explicitly defining a click emit event fixes this issue.

Only this? Or does it also fix #1129 ?

mittalyashu:
> I tested this issue as well and by explicitly defining a click emit event fixes this issue.

@Ja4V8s28Ck It specifically fixes #1129.

Can you try applying the fix on your end and confirm it as well?

Ja4V8s28Ck:
> > I tested this issue as well and by explicitly defining a click emit event fixes this issue.
> 
> @Ja4V8s28Ck It specifically fixes #1129.
> 
> Can you try applying the fix on your end and confirm it as well?

There is an emit in the `BoardSuggestion.vue`, but the issue is that the emit is named as `click` which also a function name in Vue, so Vue got confused and ran it twice. Apparently changing the name of the emit in both the parent and child resolved this issue.

packages/theme/src/pages/dashboard/posts/_slug/Index.vue
```diff
@@ -295,7 +295,9 @@ async function suggestRoadmap(event: Event) {
 function selectBoard(index: number) {
   const item = boards.suggestions[index];
 
-  Object.assign(post.board, item);
+  Object.assign(post, {
+    board: item,
+  });
```
mittalyashu, line 300: We can revert this back as well, since the `board` property in post can never be `undefined`.

Ja4V8s28Ck: 😅 I added it so that it looks similar to how roadmaps are assigned, for maintainability.

mittalyashu: ok

mittalyashu:
LGTM

PR #1137, opened 2025-09-01
add: API build depends on DB health check

Ja4V8s28Ck:
Instability affecting the production docker build has been fixed by adding DB health_check dependency for API

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -42,7 +42,8 @@ services:
       - /app/packages/server/dist
       - /app/packages/types/dist
     depends_on:
-      - db
+      db:
+        condition: service_healthy
     networks:
       - dev-network
     environment:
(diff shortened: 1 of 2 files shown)


docker-compose.dev.yml
```diff
@@ -42,7 +42,8 @@ services:
       - /app/packages/server/dist
       - /app/packages/types/dist
     depends_on:
-      - db
+      db:
+        condition: service_healthy
```
mittalyashu, line 46: Was the same issue reproducible in local development as well.

Ja4V8s28Ck: > Was the same issue reproducible in local development as well.

I haven't personally faced the issue in the local development, but still added it so that we don't get any issues in the future.

mittalyashu:
LGTM

PR #1143, opened 2025-09-02
Test(api): Filter Boards. [GET] /boards

dinirichard:
Test cases for filtering through boards.


dinirichard:
Ready for review @mittalyashu

packages/server/vitest.setup.integration.ts
mittalyashu, line 1: @dinirichard can you explain the changes in this file?

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -78,9 +274,46 @@ describe("GET /boards/search/:name", () => {
...
+        .into("permissions_roles");
+
+      // assign the role to newly created user
+      await trx
+        .insert({
+          id: uuid(),
+          role_id: newRoleId,
+          user_id: authUser.userId,
+        })
+        .into("roles_users");
```
mittalyashu: Can you use `createRoleWithPermissions` util function?

packages/server/vitest.setup.integration.ts
dinirichard: I created this file to setup a global `beforeAll()` and `afterAll()` config that runs for every parallel worker. In this case the every database inserts made in any testfile, can be saved in `globalThis.tableInserts =[]` which will all be deleted at the end of each individual worker. 

This is to prevent data collisions without using `cleanDb()`.

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -1,24 +1,213 @@
...
+  });
+
+  it("should get filtered boards in defualt filter values", async () => {
+    const filterQuery = {
+      // default page num is 0
+      // page: 0,
+      // default and max limit is 10
+      // limit: 10,
+      // defualt 'ASC' order
+      // created: "ASC",
```
mittalyashu: You have commented out the filter queries here, is that intended?

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -1,24 +1,213 @@
...
+  it("should get an Array of boards", async () => {
     const response = await supertest(app).get("/api/v1/boards");
 
     expect(response.headers["content-type"]).toContain("application/json");
     expect(response.status).toBe(200);
-    expect(response.body.boards).toHaveLength(0);
+    expect(response.body.boards).toBeInstanceOf(Array);
+  });
+
+  it("should get filtered boards in defualt filter values", async () => {
```
mittalyashu:
```suggestion
  it("should get filtered boards in default filter values", async () => {
```

packages/server/vitest.setup.integration.ts
mittalyashu: That an interesting idea 💡 

Can you create a separate PR for this change we can obsolete the `cleanDb()` all together?

mittalyashu:
@dinirichard Heads up, there were some overlapping commits and after doing a rebase seems to fixed commit history of this PR 🙂

packages/server/tests/integration/v1/boards.spec.ts
dinirichard: Yes.

dinirichard (commented):
Ready for review

mittalyashu:
@dinirichard Can you look into this failing integration test?

```
tests/integration/v1/boards.spec.ts > GET /api/v1/boards
```

dinirichard:
Ready for review.

packages/server/tests/integration/v1/boards.spec.ts
```diff
@@ -1,24 +1,197 @@
...
+interface BoardInsertRecord {
+  boardId: string;
+  name: string;
+  url: string;
+  color: string;
+  display: boolean;
+  view_voters: boolean;
+  createdAt: string;
+  updatedAt: string;
+}
```
mittalyashu, line 20: @dinirichard can you move this code to `packages/server/tests/utils/generators.ts` file and place it above the `board` function?

PR #1157, opened 2025-09-04
fix: early return in boardExists

Ja4V8s28Ck:
This is a hot-fix for the issue which allows, unintended update of a board map, when no parameters are passed.

Changes made

- `packages/server/src/ee/middleware/boardExists.ts`


mittalyashu:
Thanks for creating the patch.

PR #1165, opened 2025-09-05
Bugfix : server crash at filterpost

Ja4V8s28Ck:
**Changes made**
- Handling undefined in `packages/server/src/helpers.ts`
- Refactored and added tests in the same


packages/server/src/helpers.ts
```diff
@@ -27,7 +27,7 @@ const validUUID = (value?: string | null): string | null =>
  * Validates a list of UUIDs
  */
 function validUUIDs(value: Array<string>): string[] {
-  if (!value.length) return [];
+  if (!value?.length) return [];
```
mittalyashu: This change conflicts with the function args data type.

Ja4V8s28Ck:
```ts
function validUUIDs(value: Array<string> | null | undefined): string[]
```
Something like this then? 🤔

mittalyashu: Can't we handle this change in the `postFilter` controller itsefl?

Marking any data type `undefined` kind-of losses its strong type safety in Typescript.

Ja4V8s28Ck: As of now we use `validUUIDs` only in `postFilter`, so yeah it is possible to handle this inside `postFilter`, but if `validUUIDs` is used by many function in the future, it could be an issue where we have to manually handle all the cases. So what is your final call?

mittalyashu: _sharing the convo from Discord_

Let's not change the validUUIDs function.
That function give us a consistent and predictable pattern based on the function args and return type.
Making the args `undefined | null` will just add more ambiguity.

The linter and formatter will warning with the type mis-match as the function argument shall always pass an array of string.

for this code change even Biome linter (which we're currently using) not even throwing error. 🤔 

in according to the function args, you cannot pass any other data type to this function other than Array. 

The reason we got the server crash is because Typescript was with us till compile time.

After that we have to handle all the exceptions at the application level. 

otherwise with just type TS types we got really strong API types.

Ja4V8s28Ck: Managed to handle the error inside `filterPost.ts`. Should work now.

mittalyashu:
With regard to the change in the fallback [page numbering](https://github.com/logchimp/logchimp/pull/1165/commits/4753b3fcbcadc07687b9875dff25107fe2a1f431), there are two aspect for starting the page numbering from `1`.

1. Human POV: we always starting the page number from `1` even if the end consumer of the REST APIs is a dev
2. Technically POV: we're still starting the page number from 0 for getting the data from db

mittalyashu:
LGTM

PR #1175, opened 2025-09-06
ci(workflows): update Lighthouse test workflow to use `pull_request_target` event

mittalyashu:


Ja4V8s28Ck:
I believe it should work now.

mittalyashu:
merging it now

PR #1185, opened 2025-09-07
Refactor and included tests for votes routes

Ja4V8s28Ck:
**Changes made**
- added a new utility function `detachPermissionFromRole`
- included tests for `NOT_ENOUGH_PERMISSION` error in both  `[POST] /votes` and `[DELETE] /votes`
- added `postExists` middleware to `votes` routes


packages/server/tests/utils/detachPermissionFromRole.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { TPermission } from "@logchimp/types";
+
+type RoleInput = {
+  roleId?: string;
+  roleName?: string;
+};
+
+type RollbackInput = {
+  roleId?: string;
+  permissionId?: string;
```
mittalyashu: Question, why we're accepting `permissionId`? When we are already accepting permissions in the `detachPermissionsFromRole` function as first argument.

packages/server/tests/integration/v1/votes.spec.ts
```diff
@@ -4,36 +4,47 @@ import { v4 as uuid } from "uuid";
...
       .post("/api/v1/votes")
-      .set("Authorization", `Bearer ${authUser.authToken}`)
+      .set("Authorization", `Bearer ${user.authToken}`)
       .send({
-        postId: uuid(),
+        postId: post.postId,
       });
 
+    await rollback();
+
```
mittalyashu: Doing a `await rollback()` does not align with out data isolation goals for tests (integration + e2e).

packages/server/tests/integration/v1/votes.spec.ts
```diff
@@ -108,27 +122,38 @@ describe("POST /api/v1/votes", () => {
...
       .delete("/api/v1/votes")
-      .set("Authorization", `Bearer ${authUser.authToken}`)
+      .set("Authorization", `Bearer ${user.authToken}`)
       .send({
-        postId: uuid(),
+        postId: post.postId,
       });
 
+    await rollback();
+
```
mittalyashu: Same as above.

packages/server/tests/utils/detachPermissionFromRole.ts
Ja4V8s28Ck: We accept permissions `type: TPermission[]` as first argument in `detachPermissionFromRole`, where `TPermission` is type string of permission like `post:read`. But to re-attach the permission with a specific role, we just need the  `permissionId` rather than having the permission name and then finding the `permissionId` which is just an extra DB query.

mittalyashu:
@Ja4V8s28Ck any update on this PR?

Ja4V8s28Ck:
> @Ja4V8s28Ck any update on this PR?

I have an idea,
- Create a user
- Detach the role `@everyone` from the newly created user
- Use `createRoleWithPermissions` to create a dummy role without `vote:create`/`vote:destroy` permission and then do the tests.

By using this idea, I can safely remove `detachPermissionFromRole` utility function and the operation is isolated.

> Ready for review.

Issue #1184, opened 2025-09-06
Negative pagination input - `limit`


Ja4V8s28Ck:
**Describe the bug**
A negative number can be passed into the pagination logic, which in turn causes an issue.

**To Reproduce**
- Choose any API endpoint that takes page input. Eg. `POST /posts/get`
- Pass the value of limit as `< 0`

**Screenshots**

<img width="810" height="630" alt="Image" src="https://github.com/user-attachments/assets/0039ef34-6e4a-453d-8c4d-c1d31e571497" />

PR #1186, opened 2025-09-07
bug: fixes lower and upper bound for limit and page

### LINKED ISSUE #1183: Unbounded pagination input - `page`
**Describe the bug**
A very large page number can be passed into the pagination logic, which in turn causes an issue.

**To Reproduce**
- Choose any API endpoint that takes `page` input. Eg. `POST /posts/get`
- Pass the value of `page` as `> 999999999999999999`

**Screenshots**

Image

Ja4V8s28Ck:
Fixes #1183 and #1184

**Changes made**
- `packages/server/src/helpers.ts`
- Number of `page(s)` is capped, to avoid _BIG_INTEGER_ERROR_
- Ensures limit is a non negative number.

### DIFF
--- a/packages/server/src/helpers.ts
+++ b/packages/server/src/helpers.ts
@@ -134,13 +134,20 @@ const isDevTestEnv =
 
 function parseAndValidatePage(value?: string) {
   const n = value ? +value : NaN;
+  const PAGE_UPPER_LIMIT = 999999;
   // Default to page 1; coerce invalid/NaN/<=0 to 1
-  return !Number.isNaN(n) && n > 0 ? Math.floor(n) : 1;
+  return Math.min(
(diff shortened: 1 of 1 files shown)


mittalyashu:
I guess we need to move from offset to cursor pagination, then. 🤔

Ja4V8s28Ck:
> I guess we need to move from offset to cursor pagination, then. 🤔

In this release? or in the upcoming release?

mittalyashu:
> > I guess we need to move from offset to cursor pagination, then. 🤔
> 
> In this release? or in the upcoming release?

Apologies for the confusion.

When I saw your issue https://github.com/logchimp/logchimp/issues/1183 about page number.
For that I have said that "we need to move to cursor pagination".

PR #1218, opened 2025-09-08
Feat/more integration tests 2

mittalyashu:


packages/server/src/app.ts
mittalyashu, line 1: FYI @logchimp/intern-qa @logchimp/intern-engineer 

We're globally handling the error for decoding URI.

PR #1222, opened 2025-09-09
Docs: add pnpm installation step to setup instructions

kswaroop7777:
Description:
While setting up the project locally, I faced an issue when running npm run server:dev.
The error was:
'pnpm' is not recognized as an internal or external command
This happened because pnpm is required but was not mentioned in the setup instructions.

What I changed ?
Updated the README.md to include a step for installing pnpm globally:
npm install -g pnpm
Clarified that dependencies should be installed using pnpm install.

Why this change is useful ?
New contributors might face the same issue and not know why npm run server:dev fails.
Adding this step makes the setup process smoother and prevents confusion.


kswaroop7777:
recheck

mittalyashu:
@kswaroop7777 Please don't create duplicate PRs

https://github.com/logchimp/logchimp/pull/1223

PR #1258, opened 2025-09-14
Feat/caching with valkey

mittalyashu:


packages/server/src/services/boards/getBoardById.ts
```diff
@@ -1,23 +1,28 @@
...
 
 // utils
 import logger from "../../utils/logger";
 
 export async function getBoardById(boardId: string): Promise<IBoard | null> {
+  const cacheKey = `board:${boardId}`;
+  if (cache.isActive) {
+    const cachedBoard = await cache.valkey.get(cacheKey);
+    if (cachedBoard) {
+      return JSON.parse(cachedBoard) satisfies IBoard;
```
Ja4V8s28Ck, line 15: Considering a scenario where (_which is very unlikely_) `cachedBoard` doesn't satisfy type `IBoard`, don't we have to delete the valkey `KV` pair?

> Other than this one, the implementation is great.

mittalyashu: > Considering a scenario where (which is very unlikely) `cachedBoard` doesn't satisfy type `IBoard`

I also have few doubts on this part, as well.

> don't we have to delete the valkey KV pair?

I have set the expiration time for the key, thanks for the heads up.

mittalyashu:
LGTM, @Ja4V8s28Ck can you do a final review from your end?

Ja4V8s28Ck:
> LGTM, @Ja4V8s28Ck can you do a final review from your end?

Took me long enough to set it up my way, While eyeballing, I noticed that there was no `cache-updating`, `cache` is set when the `key` doesn't exist, but after we populate the `key-value`. The `key-value` just stays the same, there is no updating of `value`. This hinders with the intended working of `LogChimp`.

Eg, For this example we take `getBoardById`, so the reproducing steps are as follows
- Create a Board
- Create posts under that board
- Edit the name of the board
- Go to dashboard and see the posts, we can see that those posts are tagged with old name of the board.

mittalyashu:
Good catch. 👀

Ja4V8s28Ck:
> Good catch. 👀

One more thing, I would like to add. There are 2 board types we use as of now:  
- `IBoard`  
- `IBoardPrivate`  

In `boardGetById.ts`, we use `IBoard`. Assuming we are updating the cache when `updateBoard.ts` is called, `updateBoard.ts` returns data of type `IBoardPrivate`.  

In that case we should consider 2 things:  

1. We shouldn’t directly update the cache with the data returned from `updateBoard.ts`, because that would be a security risk and cause type mismatches.  
2. At some point, we will also be caching boards with type `IBoardPrivate`. Therefore, we should rename the cache key from `board:{boardId}` to something more flexible.

mittalyashu:
> One more thing, I would like to add. There are 2 board types we use as of now:
> 
> * `IBoard`
> * `IBoardPrivate`
> 
> In `boardGetById.ts`, we use `IBoard`. Assuming we are updating the cache when `updateBoard.ts` is called, `updateBoard.ts` returns data of type `IBoardPrivate`.
> 
> In that case we should consider 2 things:
> 
> 1. We shouldn’t directly update the cache with the data returned from `updateBoard.ts`, because that would be a security risk and cause type mismatches.
> 2. At some point, we will also be caching boards with type `IBoardPrivate`. Therefore, we should rename the cache key from `board:{boardId}` to something more flexible.

This is indeed a challenge we have to handle which should works for all dynamic data entities.

As a short term fix in context to this PR, I have set the cache invalidation in `update` and `delete` board operation.

Ja4V8s28Ck:
> This is indeed a challenge we have to handle which should works for all dynamic data entities.
> As a short term fix in context to this PR, I have set the cache invalidation in `update` and `delete` board operation.

LGTM, as of a long term fix, we should set the key in `updateBoard.ts` so that we don't have to reload `Dashboard/Posts` everytime we make a change in `Dasboard/Boards` to see the updated results and we can save that extra API call.

mittalyashu:
> we should set the key in `updateBoard.ts` so that we don't have to reload `Dashboard/Posts` everytime we make a change in `Dasboard/Boards` to see the updated results and we can save that extra API call.

What you have suggested is related to frontend cache and in context of this PR both Valkey cache and Frontend cache works independent of each other.

But I agree with what you've said. I've created a issue for this https://github.com/logchimp/logchimp/issues/1269.

PR #1259, opened 2025-09-14
feat: handle 404 error state for dashboard post editor page

mittalyashu:
This PR handles the dashboard post edit page when the URL provided slug is not found from API response.

### DIFF
--- a/packages/theme/src/components/dashboard/posts/PostEditor/index.vue
+++ b/packages/theme/src/components/dashboard/posts/PostEditor/index.vue
@@ -76,7 +76,7 @@
 </template>
 
 <script setup lang="ts">
-import { computed, ref } from "vue";
+import { computed, ref, reactive } from "vue";
 import type {
   IBoardPrivate,
   IDashboardPost,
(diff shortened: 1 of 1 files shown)


packages/theme/src/pages/dashboard/posts/_slug/Index.vue
```diff
@@ -1,115 +1,39 @@
...
-        </div>
-      </div>
-    </div>
+  <div
+    :class="[
+      'flex-1',
+      {
+        'flex items-center': loading || errorCode
+      }
+    ]"
```
mittalyashu, line 8: @Ja4V8s28Ck can you look at this change? Bit unsure about it.

Ja4V8s28Ck: It works well. Thing is it only works for 404. Is there a way we could write an `error page component` so that it covers all the error cases as you mentioned earlier. Something similar to the `Alert` component you wrote for the `docv2`, So that we can handle all the errors under a single component.

The newly component can take props like this
`<ErrorPage statusCode=500 statusMessage="INTERNAL_SERVER_ERROR" ORIGIN="ROADMAPS">`

mittalyashu: There are types of 404s:

1. A frontend route/page is not found and fallback to 404 page
2. A 404 status code returns from API call that needs to be handled case-by-case.

mittalyashu: I have created a `404` component: `packages/theme/src/components/dashboard/404.vue`

mittalyashu:
LGTM

PR #1281, opened 2025-09-16
Add/new uncategorized board and roadmap

Ja4V8s28Ck:
https://github.com/user-attachments/assets/3bd0832c-50f4-4239-b180-fbd98df6041a

**Changes made**
- Bug fix that disallowed `boardId` and `roadMapId` to be updated in `updatePosts` when it's `undefined`
- UI fixes
- Added a new tag called `No Board` and `No Roadmap`.


mittalyashu:
@Ja4V8s28Ck is this PR ready for review?

Ja4V8s28Ck:
> @Ja4V8s28Ck is this PR ready for review?

Yes!, after I resolve the conflicts

Ja4V8s28Ck:
> > @Ja4V8s28Ck is this PR ready for review?
> 
> Yes!, after I resolve the conflicts

Conflicts resolved, Ready for review!

packages/server/src/controllers/post/updatePost.ts
```diff
@@ -30,8 +30,8 @@ export async function updatePost(
   const id = validUUID(req.body.id);
   const title = req.body.title;
   const contentMarkdown = req.body.contentMarkdown;
-  const boardId = validUUID(req.body.boardId) || undefined;
-  const roadmapId = validUUID(req.body.roadmapId) || undefined;
+  const boardId = validUUID(req.body.boardId) || null;
+  const roadmapId = validUUID(req.body.roadmapId) || null;
```
mittalyashu: Doesn't `validUUID` function returns `null` value for invalid UUID's?

packages/theme/src/ee/components/dashboard/boards/SearchBoardDropdown/Dropdown.vue
```diff
@@ -20,7 +20,7 @@
                 <div class="text-md font-semibold line-clamp-1">
                   {{board.name}}
                 </div>
-                <p class="text-xs text-neutral-500 line-clamp-1 leading-3">
+                <p class="text-xs text-neutral-500 line-clamp-1 leading-3 pb-0.5">
```
mittalyashu: Any specific reason, you've added a padding bottom of 8px

packages/server/src/controllers/post/create.ts
```diff
@@ -31,7 +31,7 @@ export async function create(
   // @ts-expect-error
   const permissions = req.user.permissions as TPermission[];
 
-  const title = req.body.title;
+  const title = req.body.title || "new post";
```
mittalyashu, line 34: Does API throw any runtime error if we do not provide a default value?

Ja4V8s28Ck: No runtime error, I have removed `POST_TITLE_MISSING`, but in the `UI`, right after we click `Create Post`, Post is created without name and if someone saves it without a name it would be a problem, so to mitigate that and referring to `Board, Roadmap & Roles`. I used `new post` as the dummy name.

packages/server/src/controllers/post/updatePost.ts
Ja4V8s28Ck: If I remember right, `undefined` was affecting the DB query. When we unassign `board` or a `roadmap` from a post, using `undefined` to update the query will fail to update the `board` or `roadmap` to `null`. So I changed it.

packages/theme/src/ee/components/dashboard/boards/SearchBoardDropdown/Dropdown.vue
Ja4V8s28Ck: 🤔 Not sure why, maybe to keep the UI intact when `board.url` was empty (for `No board` tag)

mittalyashu: ok

packages/server/src/controllers/post/create.ts
mittalyashu: ok

PR #1295, opened 2025-09-17
feat: improve logchimp config manager

mittalyashu:
Now LogChimp reads the configuration from both `logchimp.config.json` and environment variables by taking an hybrid approach.

It reads the configuration in the following order:
1. Environment variables
2. `logchimp.config.json` file

---

This PR also clean some area of code where we were using Environment variables + configuration file in the business logic.


packages/server/src/utils/logchimpConfig.ts
```diff
@@ -1,102 +1,197 @@
...
+    return fs.existsSync(this.configPath);
+  }
+
+  // Read config from `logchimp.config.json` file
+  private getConfigFile(): Config | null {
+    if (!this.hasConfigFile()) return null;
+
+    const config = fsExtra.readJsonSync(this.configPath);
+
     if (config.mail.service) {
```
Vedant005: Optional chaining here?

docker/community/.env.example
```diff
@@ -28,6 +28,9 @@ NODE_ENV=production
 # [optional] default to port `8000`
 API_PORT=8000
 
+# [optional] default to 0.0.0.0
+LOGCHIMP_SERVER_HOST=0.0.0.0
+
 # The complete base URL at which the backend is running on.
 LOGCHIMP_API_URL="${LOGCHIMP_PROTOCOL}://localhost:${API_PORT}"
```
Vedant005, line 36: DEFAULT_SERVER_PORT not added in .env.example

packages/server/src/utils/logchimpConfig.ts
```diff
@@ -1,102 +1,197 @@
...
+
+      // Cache
+      cacheUrl: process.env.LOGCHIMP_VALKEY_URL,
+
+      // Mail
+      mailHost: process.env.LOGCHIMP_MAIL_HOST,
+      mailUser: process.env.LOGCHIMP_MAIL_USER,
+      mailPassword: process.env.LOGCHIMP_MAIL_PASSWORD,
+      mailPort: mailPort ? Number.parseInt(mailPort, 10) : 465,
+    };
```
Vedant005, line 191: What if someone give mailPort as 0? It will just use the default port right?

packages/server/src/utils/logchimpConfig.ts
```diff
@@ -1,102 +1,197 @@
...
+  databaseSsl: boolean;
+
+  // Cache
+  cacheUrl: string | undefined;
+
+  // Mail
+  mailHost: string | undefined;
+  mailUser: string | undefined;
+  mailPassword: string | undefined;
+  mailPort: number;
```
Vedant005: Why databasePort here a string but server and mail port are number?

packages/server/src/utils/logchimpConfig.ts
mittalyashu: You're saying to add optional chaining at `config.mail.service`?

docker/community/.env.example
mittalyashu: I don't see any variable `DEFAULT_SERVER_PORT` used in the codebase. 🤔

packages/server/src/utils/logchimpConfig.ts
mittalyashu: There are certain limits in which the port numbers can be set and same thing apply for the SMTP protocol as well.

Else it will fail the connection.

packages/server/src/utils/logchimpConfig.ts
Vedant005: Yes

packages/server/src/utils/logchimpConfig.ts
mittalyashu: Good point, let me check.

packages/server/src/utils/logchimpConfig.ts
mittalyashu: Hmm... let me see.

docker/community/.env.example
Vedant005: it is used here in packages/server/src/utils/logchimpConfig.ts in L117

```
 serverPort: serverPort
        ? Number.parseInt(`${serverPort}`, 10)
        : DEFAULT_SERVER_PORT,
 ````

packages/server/src/utils/logchimpConfig.ts
mittalyashu: fixed.

packages/server/src/utils/logchimpConfig.ts
mittalyashu: fixed.

docker/community/.env.example
mittalyashu: AH! That's a local variable only used to define a DEFAULT port values.

mittalyashu:
@Vedant005 Are we good to merge this PR?

Vedant005:
@mittalyashu  Yes we are good to merge

PR #1308, opened 2025-09-19
Refactor/auth verify remove request body

### LINKED ISSUE #1299: Request Body not used for `/auth/email/verify`
**Describe the bug**

The `/auth/email/verify` endpoint is defined as a POST with a request body { email: string }, but the value of email is not used. Instead, it is used from the `userExists` middleware.

**To Do**
Remove the `request body` from `/auth/email/verify` endpoint.

Ja4V8s28Ck:
Fixes #1299

**Changes made**
 - Removed unused request body from `/auth/email/verify` endpoint.

### DIFF
--- a/packages/theme/src/components/account/VerificationAlert.vue
+++ b/packages/theme/src/components/account/VerificationAlert.vue
@@ -47,7 +47,7 @@ import Button from "../ui/Button.vue";
 import { VITE_API_URL } from "../../constants";
 import { useUserStore } from "../../store/user";
 
-const { getUser } = useUserStore();
+const { authToken } = useUserStore();
 const loading = ref<boolean>(false);
 
 interface IRequestStatus {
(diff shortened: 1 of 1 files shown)


packages/server/src/routes/v1/auth.ts
```diff
@@ -18,7 +18,7 @@ router.post("/auth/setup", mailConfigExists, auth.setup);
 router.get("/auth/setup", auth.isSiteSetup);
 
 // email
-router.post(
+router.get(
```
mittalyashu: Keep the HTTP method to POST for now.

Ja4V8s28Ck: For now? 👀 Should I put a hold on that issue? Why though, I haven't seen any breaking changes as of now.

mittalyashu: If a customer tries to make an API call using `POST` they would get a 404, because of this change.

mittalyashu:
LGTM

PR #1311, opened 2025-09-19
fix(workflow) trufflehog

Ja4V8s28Ck:
> Test playground to make sure trufflehog runs as intended


.config/.trufflehogignore
```diff
@@ -0,0 +1,2 @@
+.git/config
```
mittalyashu, line 1: Question, don't we need to ignore the whole `.git` directory

mittalyashu (changes requested):

.config/.trufflehogignore
Ja4V8s28Ck: Because, I wanted to minimized area of ignorance, that's why I only added `.git/config`. During `Checkout` job in CI, an auth token is added by `Checkout` and `Trufflehog` catches it.

mittalyashu: Hmm... 🧐

Ja4V8s28Ck: One more thing, as far as the scope of this PR, this workflow runs as intended. As of how it would work during a merge is different story.

mittalyashu: _let's see how it goes_.

mittalyashu:
Merging this PR for now, let's see how it goes.

PR #1316, opened 2025-09-21
Refactor(board): check name to check slug

Ja4V8s28Ck:
- Renamed `checkName` → `checkSlug` and it's respective affected areas.
- Board URL is now validated for availability before update.
- Added Debounce for checking board url availability.
- Restricted nanoid generation only to [a-z0-9_-].
- `checkSlug` will also take `board:update` as permission.
- Added integration tests for `check-slug` and update board.
- Added `checkName` for backward compatibility.

### DIFF
--- a/packages/server/src/ee/controllers/v1/boards/checkSlug.ts
+++ b/packages/server/src/ee/controllers/v1/boards/checkSlug.ts
@@ -1,5 +1,5 @@
 import type { Request, Response } from "express";
-import type { TBoardCheckNameBody, TPermission } from "@logchimp/types";
+import type { TBoardCheckSlugBody, TPermission } from "@logchimp/types";
 import database from "../../../../database";
 
 // utils
@@ -7,13 +7,13 @@ import logger from "../../../../utils/logger";
 import error from "../../../../errorResponse.json";
 
 export async function checkSlug(
-  req: Request<unknown, unknown, TBoardCheckNameBody>,
+  req: Request<unknown, unknown, TBoardCheckSlugBody>,
   res: Response,
 ) {
   // @ts-expect-error
   const permissions = req.user.permissions as TPermission[];
 
-  const name = req.body.name;
+  const slugUrl = req.body.url;
 
   const checkPermission = permissions.includes("board:create");
   if (!checkPermission) {
@@ -24,15 +24,15 @@ export async function checkSlug(
     return;
   }
 
-  if (!name) {
+  if (!slugUrl) {
     res.status(400).send({
-      message: error.api.boards.nameMissing,
-      code: "BOARD_NAME_MISSING",
+      message: error.api.boards.urlMissing,
+      code: "BOARD_URL_MISSING",
     });
     return;
   }
 
-  const slimUrl = name
+  const slimUrl = slugUrl
     .replace(/[^\w]+/gi, "-")
     .trim()
     .toLowerCase();
--- a/packages/server/src/ee/controllers/v1/boards/checkSlug.ts
+++ b/packages/server/src/ee/controllers/v1/boards/checkSlug.ts
@@ -6,7 +6,7 @@ import database from "../../../../database";
 import logger from "../../../../utils/logger";
 import error from "../../../../errorResponse.json";
 
-export async function checkName(
+export async function checkSlug(
   req: Request<unknown, unknown, TBoardCheckNameBody>,
   res: Response,
 ) {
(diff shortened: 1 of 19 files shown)


Ja4V8s28Ck:
Ready for review.

packages/server/src/ee/controllers/v1/boards/create.ts
```diff
@@ -36,6 +36,7 @@ export async function create(
   }
 
   try {
+    const nanoid = customAlphabet("_0123456789abcdefghijklmnopqrstuvwxyz");
```
mittalyashu: better to move this to helpers file

packages/server/src/ee/controllers/v1/boards/checkSlug.ts
```diff
@@ -1,21 +1,23 @@
...
-  req: Request<unknown, unknown, TBoardCheckNameBody>,
+export async function checkSlug(
+  req: Request<unknown, unknown, TBoardCheckSlugBody>,
   res: Response,
 ) {
   // @ts-expect-error
   const permissions = req.user.permissions as TPermission[];
 
-  const name = req.body.name;
+  const slugUrl = req.body.url;
```
mittalyashu, line 23: Keep `req.body.name` for backward compatibility.

And add a log depreciation warning if `req.body.name` property is provided by the user.

_You can take reference from `[GET] /api/v1/users` API endpoint._

packages/server/src/ee/controllers/v1/boards/checkSlug.ts
```diff
@@ -24,15 +26,15 @@ export async function checkName(
     return;
   }
 
-  if (!name) {
+  if (!slugUrl) {
     res.status(400).send({
-      message: error.api.boards.nameMissing,
-      code: "BOARD_NAME_MISSING",
+      message: error.api.boards.urlMissing,
+      code: "BOARD_URL_MISSING",
```
mittalyashu, line 39: Keep the same error code? For backward compatibility.

packages/server/src/ee/routes/v1/boards.ts
```diff
@@ -34,7 +34,7 @@ router.get<ISearchBoardRequestParams>(
   boards.searchBoard,
 );
 
-router.post("/boards/check-name", authRequired, boards.checkName);
+router.post("/boards/check-slug", authRequired, boards.checkSlug);
```
mittalyashu, line 37: That's definitely a breaking change.

packages/server/src/ee/controllers/v1/boards/checkSlug.ts
Ja4V8s28Ck: I thought of that, but given that only we know about all possible API calls (_Lack of API documentation, so users are only exposed to UI capabilities of LogChimp and only a handful of advanced users could have used the API capabilities_). This would be the best time to change API's without any backward compatibility.

packages/server/src/ee/routes/v1/boards.ts
Ja4V8s28Ck: Considering, It's only used in `Board Settings`, the scope for the breaking is narrow.

packages/server/src/ee/controllers/v1/boards/checkSlug.ts
Ja4V8s28Ck: Personally, it's very confusing as `Board` already has a `name` that's referred as `title`.

packages/server/src/ee/controllers/v1/boards/create.ts
Ja4V8s28Ck: On it

packages/server/src/ee/controllers/v1/boards/checkSlug.ts
mittalyashu: > I thought of that, but given that only we know about all possible API calls (Lack of API documentation, so users are only exposed to UI capabilities of LogChimp

You're right about that. 😅

---

We following semantic versioning, making this change means doing major release `v1.0.0`

What you can do, is something like this:

```ts
 const name = req.body.name || req.body.url;
 ```

Update the `checkName` API endpoints type definitions and add a log depreciation warning if `req.body.name` property is provided by the user.

_You can take reference from `[GET] /api/v1/users` API endpoint._

Ja4V8s28Ck:
Ready for review.

mittalyashu:
_rebased the PR_

mittalyashu:
LGTM

PR #1334, opened 2025-09-25
fix: trigger posts load on small screens#1325

### LINKED ISSUE #1325: Infinite scroll not working on small screen height
**Screenshots**

| Normal screen height | Smaller screen height |
| --- | --- |
| Image | Image |

aavizmehar:
Fixes #1325
Posts were not loading automatically on small screens.
Problem with window scroll
trigger loading based on window scroll, it only fires when the page is actually scrollable.
On small screens, the page may not have enough height to scroll → infinite scroll never triggers.

Solution:
placed a small <div>"infiniteTrigger" at the bottom of the list.
Use useInfiniteScroll on "infiniteTrigger" to detect when it comes into view.
When the element enters the viewport load more items.

## Summary by CodeRabbit

* **New Features**
  * Infinite scroll now uses an invisible page trigger element for more reliable detection.
  * Auto-loads content when the page isn’t tall enough to require scrolling.
  * Added a “Completed” state to indicate no more results and stop further loads.

* **Bug Fixes**
  * Initial results now load automatically instead of waiting for user scroll.

* **Style**
  * Minor formatting adjustment with no functional impact.

### DIFF
--- a/packages/theme/src/components/ui/InfiniteScroll.vue
+++ b/packages/theme/src/components/ui/InfiniteScroll.vue
@@ -100,7 +100,7 @@ onMounted(async () => {
   const viewportHeight = window.innerHeight;
   const rect = triggerEl.getBoundingClientRect();
 
-  if (rect.top < viewportHeight) {
+  if (viewportHeight < 200 || rect.top < viewportHeight) {
     executeInfiniteScroll();
   }
 });


packages/theme/src/pages/Index.vue
```diff
@@ -76,8 +74,15 @@ async function loadMorePosts() {
...
-onMounted(() => isSetup());
+onMounted(() => {
+  isSetup();
+  nextTick(() => {
+    // if smallscreen, loadmoreposts
+    if (window.innerHeight < 524) {
+      loadMorePosts();
+    }
+  });
+});
```
mittalyashu: @aavizmehar Before creating a quick patch, can you look into the root cause of this issue.

aavizmehar: the root cause is that useInfiniteScroll is triggered automatically when a scroll event happens but on small viewports with minimal content, that scroll event never happens.
 so we need to force the initial data load on component mount.

mittalyashu: If the issue is occurring in the infinite scroll logic and we're are using the same infinite scroll logic at various places.
It would be better to apply the fix there.

aavizmehar: placed a small <div>"infiniteTrigger" at the bottom of the list.
Use useInfiniteScroll on "infiniteTrigger" to detect when it comes into view.
When the element enters the viewport load more items.

Vedant005:
@aavizmehar solution does not work for screen height < 185px

mittalyashu:
@aavizmehar Can you check on the comments shared by the CodeRabbit bot?

- Reply to the comment if you think otherwise,
- Or, mark it as resolved.

mittalyashu:
@aavizmehar Biome CI pipeline is failing.

mittalyashu:
@aavizmehar there seems to be some code formatting issues.

packages/theme/src/components/ui/InfiniteScroll.vue
```diff
@@ -18,17 +18,18 @@
         </client-error>
       </slot>
     </template>
+      <div ref="infiniteTrigger" aria-hidden="true" style="height:1px;" />
```
mittalyashu, line 21: @aavizmehar I can see you've added this line.

Did you checked what UI changes this `div` element does to the design?

aavizmehar: the 1px height is needed for the observer to detect the trigger and is barely noticeable in the ui.

aavizmehar:
@mittalyashu Ready for review

mittalyashu:
@aavizmehar In a PR, instead of doing a git merge, please do git rebase.

CleanShot 2025-10-22 at 15 03 18@2x

Issue #1342, opened 2025-09-27
Frontend continue to make API call on network mode set to offline


mittalyashu:
The logic continues to make API call on offline network, this can greatly impact the performance or crash the application itself.

The issue does not seems specific t roadmaps page but at all the places where infinite scroll is being used.

![Image](https://github.com/user-attachments/assets/439f3d24-f7ec-4051-980d-9b4cc6239c34)

D0R4GON:
Hi, I would like to work on this issue... Could you assign it to me?

mittalyashu:
@D0R4GON You can work on it.

PR #1344, opened 2025-09-27
added a check, if offline: oninfinite() not called    #1342

aavizmehar:
Previously, the infinite scroll component would continue calling onInfinite() even when the user was offline. This lead to unnecessary API errors, performance issues, or potential crashes.

Now, the component checks navigator.onLine before triggering onInfinite(). If the network is offline, no API calls are made, preventing repeated errors and improving stability.

This change applies to all places where infinite scroll is used, not just the roadmaps page

### DIFF
--- a/packages/theme/src/components/ui/InfiniteScroll.vue
+++ b/packages/theme/src/components/ui/InfiniteScroll.vue
@@ -76,6 +76,7 @@ watch(
 
 function executeInfiniteScroll() {
   if (props.state === "COMPLETED" || props.state === "ERROR") return;
+    if (!navigator.onLine) return;
   props.onInfinite();
 }
 


mittalyashu:
@aavizmehar It is still making API calls even after making the network status to offline.


https://cln.sh/H6nJdP9yGR47jMz05fPJ

aavizmehar:
added a check in the canLoadMore so it returns false if offline so infinite scroll is not triggered.

packages/theme/src/pages/Roadmaps.vue
```diff
@@ -65,7 +65,7 @@ useInfiniteScroll(
   {
     direction: "right",
     canLoadMore: () => {
-      return hasNextPage.value;
+      return hasNextPage.value && navigator.onLine;
```
mittalyashu, line 68: @aavizmehar The fix you've applied would only works in the roadmaps page.

What about other places where infinite scroll is being used?

mittalyashu:
@aavizmehar The offline behaviour seems to be in-consistent across different pages.

This screencast is from dashboard boards page.



https://github.com/user-attachments/assets/0e0563ca-9273-4f2e-b09a-eaac8f71717e

mittalyashu:
@aavizmehar Can you link the Github issue this PR closes in the PR description?

aavizmehar:
@mittalyashu  Ready for review

packages/theme/src/components/ui/InfiniteScroll.vue
```diff
@@ -82,6 +82,10 @@ function executeInfiniteScroll() {
 useInfiniteScroll(window, executeInfiniteScroll, {
   distance: props.distance,
   direction: "bottom",
-  canLoadMore: () => !noMoreResults.value || props.state !== "ERROR",
+  canLoadMore: () =>
+    navigator.onLine &&
+    props.canLoadMore &&
+    !noMoreResults.value &&
+    props.state !== "ERROR",
```
mittalyashu, line 89: @aavizmehar Any specific reason you changed from `!noMoreResults.value || props.state !== "ERROR"` to `!noMoreResults.value && props.state !== "ERROR"`?

mittalyashu (changes requested):

packages/theme/src/components/ui/InfiniteScroll.vue
Aaviz1: Critical Logic Bug: The OR operator (||) is incorrect. When noMoreResults.value === true and props.state === "COMPLETED", the function incorrectly returns true because "COMPLETED" !== "ERROR" evaluates to true. This allows infinite scroll to attempt loading when no more results are available

mittalyashu: > Critical Logic Bug: The OR operator (||) is incorrect. When noMoreResults.value === true and props.state === "COMPLETED", the function incorrectly returns true because "COMPLETED" !== "ERROR" evaluates to true. This allows infinite scroll to attempt loading when no more results are available

This is what the AI says, have you tested this scenario manually?

mittalyashu: //cc @aavizmehar

PR #1318, opened 2025-09-23
fixed site update with empty siteName #1312

mrayushmehrotra:
added a return statement after the siteName.value and accentColor.value check so that function will not proceed further. and fix issue #1312

PR #1356, opened 2025-10-05
refactor to better approach

mrayushmehrotra:


mittalyashu:
@mrayushmehrotra Have you tested the code change in this PR?

packages/theme/src/pages/dashboard/settings/General.vue
```diff
@@ -202,7 +202,6 @@ async function updateSettingsHandler() {
       siteName.error.show = true;
       siteName.error.message = "Required";
     }
-
     if (!accentColor.value) {
       accentColor.error.show = true;
       accentColor.error.message = "Required";
```
TaranaKhanna, line 207: here if accentColor and sitname.value not provided, the error are given but this does not prevent the function from calling: const response = await updateSettings(siteData);

mittalyashu: Yes, and @mrayushmehrotra has created a PR to fix that.

packages/theme/src/pages/dashboard/settings/General.vue
TaranaKhanna, line 1: inside updateSettingsHandler function after checking for siteName.value && accentColor.value you show error but it does not stop the function. even if these siteName.value && accentColor.value are not present, 
updateSettingsButtonLoading.value is set to true and then you again check for the same inside try block using: if (siteName.value && accentColor.value), and then the catch block runs which is not needed. we can stop the function right after the if statement using return.

mittalyashu: @TaranaKhanna The author of the PR tried attempting something similar in [different PR](https://github.com/logchimp/logchimp/pull/1318/files).

That didn't worked either.

packages/theme/src/pages/dashboard/settings/General.vue
```diff
@@ -222,16 +221,16 @@ async function updateSettingsHandler() {
...
-
     siteName.value = response.data.settings.title;
     logo.value = response.data.settings.logo;
     description.value = response.data.settings.description;
     accentColor.value = response.data.settings.accentColor;
     googleAnalyticsId.value = response.data.settings.googleAnalyticsId;
     developer_mode.value = response.data.settings.developer_mode;
-
     update(response.data.settings);
+    }
```
mittalyashu, line 233: @coderabbitai what do you have to say about this change?

packages/theme/src/pages/dashboard/settings/General.vue
TaranaKhanna: ok I have not ran it, I will check.

mittalyashu: @TaranaKhanna no update since your last comment?

mittalyashu:
Closing this PR due to author inactivity.

PR #1366, opened 2025-10-07
fix/roadmap dropdown reactivity

Ja4V8s28Ck:
Fixes the issue where the roadmap `MoreOptionDropdowns` is stale even if the roadmap gets deleted or if the idx changes.

EDIT: _I DIDN'T DO ANYTHING ABOUT CODERABBIT_

## Summary by CodeRabbit

* **New Features**
  * New roadmaps are now appended to the end of the list, preserving existing ordering.

* **Bug Fixes**
  * Dropdown options correctly disable when no roadmap is available.
  * Roadmap-dependent UI now updates live when roadmap data changes, reducing stale displays.

* **Style**
  * Minor formatting cleanup.

### DIFF
--- a/packages/theme/src/ee/components/dashboard/roadmap/TabularItem/options.ts
+++ b/packages/theme/src/ee/components/dashboard/roadmap/TabularItem/options.ts
@@ -1,7 +1,9 @@
-import type { InjectionKey } from "vue";
+import type { InjectionKey, ComputedRef } from "vue";
 import type { IRoadmapPrivate } from "@logchimp/types";
 
 /**
  * Symbol to exchange roadmap's data
  */
-export const roadmapKey = Symbol() as InjectionKey<IRoadmapPrivate>;
+export const roadmapKey = Symbol() as InjectionKey<
+  ComputedRef<IRoadmapPrivate>
+>;


packages/theme/src/ee/components/dashboard/roadmap/TabularItem/TabularItem.vue
```diff
@@ -69,7 +69,10 @@ interface Props {
   roadmap: IRoadmapPrivate;
 }
 const props = defineProps<Props>();
-provide(roadmapKey, props.roadmap);
+provide(
+  roadmapKey,
+  computed(() => props.roadmap),
```
mittalyashu, line 74: According to the Vue.js docs, providing a `ref` (reactive) value to `provide` should keep the value reactive without wrapping the value in `computed()`.

Ja4V8s28Ck: Using `ref` works only if we watch for change in the props and update it. Using `computed` aligns more closely with a cleaner solution. If we use the `ref` would change into something like this.
```ts
const reactiveRoadmap = ref(props.roadmap);
provide(roadmapKey, reactiveRoadmap);
watch(
  () => props.roadmap,
  (newVal) => {
    reactiveRoadmap.value = newVal;
  },
);
```

mittalyashu:
Closed by https://github.com/logchimp/logchimp/pull/1378

PR #1418, opened 2025-10-20
Fix : Viewer Vote status can now be retrieved only by the authenticated/logged in user

### LINKED ISSUE #1321: Vote status can be retrieved for any user by providing `userId`

Vedant005:
Fixes  #1321 

Changes :

- Sent authToken to backend for validation
- Applied authRequired middleware to validate the token and get user details
- Used the userId received from the middleware to fetch viewer vote status of only the logged in user

## Summary by CodeRabbit

* **Bug Fixes**
  * Post retrieval now requires authentication; clients send an authorization token instead of including a userId in the request.
  * Voting and access checks now rely on authenticated identity, improving vote attribution and permission enforcement.

* **Tests**
  * Added comprehensive integration tests covering auth header errors, invalid/empty tokens, missing user, blocked users, and missing permissions.

### DIFF
--- a/packages/types/src/posts/post.ts
+++ b/packages/types/src/posts/post.ts
@@ -25,7 +25,6 @@ interface IPostInfo {
 }
 
 export interface IFilterPostRequestBody {
-  userId?: string;
   boardId: string[];
   roadmapId?: string;
   page: string;


packages/server/src/routes/v1/posts.ts
```diff
@@ -9,7 +9,7 @@ import * as post from "../../controllers/post";
 import { authRequired } from "../../middlewares/auth";
 import { postExists } from "../../middlewares/postExists";
 
-router.post("/posts/get", post.filterPost);
+router.post("/posts/get", authRequired, post.filterPost);
```
mittalyashu, line 12: @Vedant005 Did you tested this change in the LogChimp Theme?

packages/server/src/controllers/post/filterPost.ts
mittalyashu, line 1: @Vedant005 You have not updated the `IFilterPostRequestBody` interface.

packages/server/src/routes/v1/posts.ts
Vedant005: Yes

packages/server/src/controllers/post/filterPost.ts
Vedant005: Done. Updated now.

packages/theme/src/modules/posts.ts
mittalyashu: We no longer need pass this property.

packages/server/src/routes/v1/posts.ts
mittalyashu, line 12: @Vedant005 Can you re-evaluate this change again?

packages/theme/src/modules/posts.ts
Vedant005: I have removed that

packages/server/src/routes/v1/posts.ts
Vedant005: @mittalyashu  
Well other way to solve this issue would be to keep userId in body but verify it against authentication.

Wait now that I look into it I can use authOptional middleware so that loggedout user can view votes.

PR #1427, opened 2025-10-22
fix: Create unique username with nanoid to avoid violation of unique constraint "users_username_unique"

### LINKED ISSUE #1319: Handle `duplicate key value violates unique constraint "users_username_unique"` durning signup
**Steps to reproduce**

1. Signup with email `new1@example.com`
2. _logout_
3. Signup with email `new1@e.com`

Vedant005:
Fixes #1319 

Used nanid and already existing helper function to create unique username

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved username generation during registration: usernames are now based on the email prefix, automatically truncated as needed and given a short unique suffix to ensure uniqueness and avoid conflicts.
  * Updated tests to stop asserting exact username values and instead validate account email and status.

### DIFF
--- a/packages/server/tests/integration/v1/auth/signup.spec.ts
+++ b/packages/server/tests/integration/v1/auth/signup.spec.ts
@@ -26,8 +26,6 @@ describe("POST /api/v1/auth/signup", () => {
 
   it("should create new user", async () => {
     const randomEmail = faker.internet.email();
-    const username = randomEmail.split("@")[0];
-
     const response = await supertest(app).post("/api/v1/auth/signup").send({
       email: randomEmail,
       password: "password",
@@ -35,7 +33,6 @@ describe("POST /api/v1/auth/signup", () => {
 
     const user = response.body.user;
     expect(response.status).toBe(201);
-    expect(user.username).toBe(username);
     expect(user.email).toBe(randomEmail.toLowerCase());
   });
 


packages/server/src/services/auth/createUser.ts
```diff
@@ -149,3 +157,16 @@ const createUser = async (
 };
 
 export { createUser };
+
+async function generateUniqueUsername(baseUsername: string): Promise<string> {
+  let username = baseUsername;
+  let exists = await database("users").where({ username }).first();
+
+  while (exists) {
```
mittalyashu: Using `while` loop, is a big no go.

Loop can trigger multiple sequential database queries.

---

Check with nanoid docs at how much length the random string has the change to get repeated and accordingly try to increase/decease the char length of random string.

mittalyashu: @coderabbitai anything you would like to share?

packages/server/src/services/auth/createUser.ts
```diff
@@ -51,7 +55,9 @@ const createUser = async (
   const name = sanitiseName(userData.name);
 
   // get username from email address after truncating to first 30 characters and sanitise
-  const username = sanitiseUsername(userData.email.split("@")[0].slice(0, 30));
+  const baseUsername = sanitiseUsername(
+    userData.email.split("@")[0].slice(0, 30),
+  );
```
mittalyashu, line 60: > This is rough idea, you can further improve it to handle any edge-cases.

@Vedant005 What's your take on this approach? 👇

You directly pass the split-ed raw username to the function.

And based on the length of the `baseUsername`, if length >= 22 then we get the first 21 char of base username and append with `_{nanoid}`, else `{username}_{nanoid}`.

---

The reason because as per your current implementation in `generateUniqueUsername` function you slice the username after append the nanoid.

packages/server/src/services/auth/createUser.ts
```diff
@@ -149,3 +157,12 @@ const createUser = async (
 };
 
 export { createUser };
+
+function generateUniqueUsername(baseUsername: string): string {
+  let username = baseUsername;
+
+  const suffix = nanoid(8);
+  username = `${baseUsername}-${suffix}`.slice(0, 30);
```
mittalyashu: Can you double-check? I think we don't support `-` in the username as per `sanitiseUsername` function.

mittalyashu (changes requested):

mittalyashu:
~~Alright, let me continue with this PR to move faster 🙂~~

Vedant005:
> Alright, let me continue with this PR to move faster 🙂

I was working on this now. I'll work on other ones then

mittalyashu:
> > Alright, let me continue with this PR to move faster 🙂
> 
> I was working on this now. I'll work on other ones then

If you're working on it, you can continue working on it. 🙂

packages/server/src/services/auth/createUser.ts
Vedant005: Well, it creates username with `-`

`Joe-n9j4c0fh`

Vedant005: I'll fix it

packages/server/src/services/auth/createUser.ts
```diff
@@ -149,3 +157,18 @@ const createUser = async (
...
+  let trimmedBase: string;
+
+  if (baseUsername.length >= 22) {
+    trimmedBase = baseUsername.slice(0, 21);
+  } else {
+    trimmedBase = baseUsername;
+  }
+
+  return `${trimmedBase}${separator}${suffix}`;
+}
```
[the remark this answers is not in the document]
Vedant005: @mittalyashu

Vedant005:
@mittalyashu  made changes. Check if proper

packages/server/src/services/auth/createUser.ts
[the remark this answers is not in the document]
mittalyashu: [This PR](https://github.com/logchimp/logchimp/pull/1494) should handle this case by further decreasing the probability of conflicting username in the database.

PR #1462, opened 2025-11-03
fix: handle timestamp collisions in user cursor pagination

Ja4V8s28Ck:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cursor-based pagination for user lists to correctly handle composite cursors and both ascending/descending sort directions.
  * Ensured metadata and page calculations use the selected sort direction so total pages, current page and next-page detection remain consistent.

### DIFF
--- a/packages/server/src/services/users/getUsers.ts
+++ b/packages/server/src/services/users/getUsers.ts
@@ -15,6 +15,7 @@ interface GetUserQueryOptions {
 
 interface GetUserMetadataOptions {
   after?: string;
+  created: ApiSortType;
 }
 
 export async function getUsers({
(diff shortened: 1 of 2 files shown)


packages/server/src/services/users/getUsers.ts
mittalyashu, line 1: @Ja4V8s28Ck You can take a more simplest approach as I have done in roles pagination PR: https://github.com/logchimp/logchimp/pull/1456/files

Ja4V8s28Ck: I read your implementation, but it didn't have `created` param which means a lot of raw query is needed, also I'm trying to find a way to avoid raw query.

mittalyashu: >  but it didn't have created param which means a lot of raw query is needed

wdym, it didn't have `created` param?

Ja4V8s28Ck: The cursor parameter `created` (`DESC` or `ASC`). It's missing. The role ordering is hardcoded   to be descending. 

image

Ja4V8s28Ck:
Ready for review!

PR #1463, opened 2025-11-05
fix(api): eliminate race condition in user signup causing flaky USER_EXISTS test

MattHandzel:
Remove non-atomic check-then-insert pattern that allowed race condition
between duplicate user checks. Now relies on database unique constraint
for atomicity, catching PostgreSQL error 23505 for duplicate emails.

Fixes intermittent test failures in signup.spec.ts where USER_EXISTS
check would occasionally return 201 instead of 409.

I ran this 20 times locally and it was successful 😋

## Summary by CodeRabbit

* **Refactor**
  * Optimized user creation by leveraging database constraints for duplicate email detection instead of preliminary existence checks, with improved error handling for constraint violations.

### DIFF
--- a/packages/server/src/services/auth/createUser.ts
+++ b/packages/server/src/services/auth/createUser.ts
@@ -61,29 +61,8 @@ const createUser = async (
   const hashedPassword = hashPassword(userData.password);
 
   try {
-    const {
-      rows: [getUser],
-    } = await database.raw(
-      `
-        SELECT EXISTS (
-          SELECT * FROM users WHERE LOWER(email) = LOWER(:email)
-        )
-      `,
-      {
-        email,
-      },
-    );
-
-    const userExists = getUser.exists;
-    if (userExists) {
-      res.status(409).send({
-        message: error.middleware.user.userExists,
-        code: "USER_EXISTS",
-      });
-      return null;
-    }
-
     // insert user to database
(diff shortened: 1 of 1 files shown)


MattHandzel:
I have read the CLA Document and I hereby sign the CLA

Ja4V8s28Ck:
> Remove non-atomic check-then-insert pattern that allowed race condition between duplicate user checks. Now relies on database unique constraint for atomicity, catching PostgreSQL error 23505 for duplicate emails.

By making the `check-then-insert` operation into `insert-catch-error` pattern, this PR addresses the concurrency issue, but not the race condition. The race condition comes from the function `cleanDB()` and parallelized file testing in vitest. Although, this change would reduce the rate of error. https://github.com/logchimp/logchimp/blob/413996e6cf455bdeb254dfb23735490f97626111/packages/server/tests/integration/v1/auth/setup.spec.ts#L9

Also one of the flaws is that, _username unique constraint_ is not handled. As the below code shows, we create the username based on the email **(as of now)**. This PR #1427 should be resolving the issue. But for the time being it would be nice to handle _username unique constraint_.
https://github.com/logchimp/logchimp/blob/413996e6cf455bdeb254dfb23735490f97626111/packages/server/src/services/auth/createUser.ts#L53-L54

Test still fails. So yeah the flakiness is not yet resolved.
image

PR #1537, opened 2026-01-07
feat: auto add prerequisite permission while enabling permission

### LINKED ISSUE #1110: Auto enable prerequisite permission for manually enabled permission
For now all the permissions can be manually toggled based on the requirement. This surface an potential issue of chronological order.

**Example 1:**

1. Owner create a new role
2. Only enabled `board:destroy` permission.
3. Assign the role to `Person1`
4. `Person1` login go to dashboard > boards page
5. Sees an error "Something went wrong"

As you can see in the above example, the `Person1` only had a `board:destroy` permission enabled so technically the `Person1` should not be able to read the data and only allow to delete.

That being said, we should add option in the frontend that the **prerequisite permission** should gets auto enabled for the respective role.

Continuing the **Example 1** further.

1. Owner create a new role
2. Owner enabled `board:destroy` permission.
3. `board:read` permission gets auto enabled.
4. Assign the role to `Person1`
5. `Person1` login go to dashboard > boards page
6. Sees list of boards

Ja4V8s28Ck:
Fixes #1110 

**Changes made**
- Added auto add permission prerequisite based on the following diagram.
image

> [!WARNING]
> `post:read` permission should be added to `@everyone` role as it already has `post:create` permission, which requires read access as a prerequisite. However, this is not updated in this PR as it's a breaking change.

## Summary by CodeRabbit

* **New Features**
  * Role name and description are editable directly in the role editor.
  * Permissions now enforce prerequisites across all sections (Posts, Boards, Roadmaps, Vote, Dashboard, Roles, Settings): enabling a permission auto-enables required dependencies; disabling revokes dependents.
  * Toggle controls immediately update related permissions when changed for consistent, real-time behavior.

### DIFF
--- a/packages/theme/src/ee/components/roles/RoleEditor.vue
+++ b/packages/theme/src/ee/components/roles/RoleEditor.vue
@@ -281,7 +281,12 @@
 
 <script setup lang="ts">
 import { computed, reactive, ref } from "vue";
-import type { IRole, TPermission, IPermissionsState } from "@logchimp/types";
+import type {
+  IRole,
+  TPermission,
+  PermissionAction,
+  IPermissionsState,
+} from "@logchimp/types";
 import { ShieldAlert } from "lucide-vue";
 
 // modules
@@ -320,7 +325,13 @@ const updateRoleButtonDisabled = computed(() => {
   return !checkPermission;
 });
 
-const permissionPrerequisite = {
+type PermissionPrerequisiteMap = {
+  [K in keyof IPermissionsState]: Partial<{
+    [action: string]: PermissionAction[];
+  }>;
+};
+
+const permissionPrerequisite: PermissionPrerequisiteMap = {
   post: {
     create: ["read"],
     update: ["read"],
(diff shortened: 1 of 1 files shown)


mittalyashu:
@Ja4V8s28Ck I skimmed through the code, from what I understand you have set the `*:read` permission as prerequisite for all other actions (create, update, delete).

You have not taken the `dashboard:read` permission into account.

I still think this will be bit tricky to handle. Or we might need to take a whole different approach as to how we handle roles and permissions when it comes to prerequisite permissions.

The `dashboard:read` permission is mainly used for Theme purposes and without the user will not able access the dashboard to perform the action in the first place.

Ja4V8s28Ck:
> You have not taken the `dashboard:read` permission into account.

Oh shoot, you are right. I'll look into this.

Ja4V8s28Ck:
@mittalyashu So I've been thinking about pros and cons of including `dashboard:read`. I feel .

> @Ja4V8s28Ck I skimmed through the code, from what I understand you have set the `*:read` permission as prerequisite for all other actions (create, update, delete).
> 
> You have not taken the `dashboard:read` permission into account.
> 
> I still think this will be bit tricky to handle. Or we might need to take a whole different approach as to how we handle roles and permissions when it comes to prerequisite permissions.
> 
> The `dashboard:read` permission is mainly used for Theme purposes and without the user will not able access the dashboard to perform the action in the first place.

So, I took note of the comments you made and the did a bit of skimming on my end. It is really tricky to handle `dashboard` situation. 
- If a user is given just the `dashboard:read` permission they are able to see
	- Boards (public & private)
	- Roadmaps (public & private)
	- Billing data
	- Site settings
Ofc the user can't modify them, still it breaks point of having RBAC.

PR #1552, opened 2026-02-13
Cached `GET /auth/setup` endpoint.

Ja4V8s28Ck:
> I haven't followed the conventional method of caching and invalidating for this endpoint. As the site is setup only once, I set the cache only when site is setup successfully and didn't bother to invalidate the cache.

## Summary by CodeRabbit

* **Refactor**
  * Optimized the site-setup status endpoint with a caching layer for faster, more consistent responses; responses will indicate cache hits/misses and cache failures are handled gracefully to avoid user-facing errors.


mittalyashu:
LGTM

PR #1573, opened 2026-02-20
fix(server/boards): add permission checks for accessing private boards

mittalyashu:
## Summary by CodeRabbit

* **New Features**
  * Board visibility is now permission-gated: users without read permission see only publicly displayed boards.
  * Board fetch endpoints now allow optional authentication so public boards remain accessible unauthenticated.

* **Bug Fixes**
  * Board generator/creation respects explicit public/private flags (false preserved).
  * Missing or invalid board requests now return appropriate permission/validation responses.

* **Tests**
  * Added/updated tests for unauthenticated and permissioned access to public and private boards.


mittalyashu:
LGTM

PR #1461, opened 2025-11-01
Fix: reset email branding

Vedant005:
Closes #1457 

image

## Summary by CodeRabbit

* **New Features**
  * Password reset emails now include your site's configured accent color and logo for consistent branding.

* **Improvements**
  * Email templates use a site color variable with a sensible fallback and display the configured logo with accessible alt text for better visual consistency and accessibility.

PR #1584, opened 2026-02-27
Fix: reset email branding - Part 2

mittalyashu:
This PR is extension of https://github.com/logchimp/logchimp/pull/1461

## Summary by CodeRabbit

* **New Features**
  * Emails (verification & password reset) now use your site's accent color and logo dynamically with sensible fallbacks; action button color adapts to branding.

* **Accessibility**
  * Logo images in emails include descriptive alt text.

* **Bug Fixes**
  * Improved handling when site branding is missing to ensure consistent email appearance.

* **Tests**
  * Added tests verifying HTML/text emails include dynamic branding and links.

* **Chores**
  * Expanded unit test discovery and added a fallback brand color constant.

### DIFF
--- a/packages/server/vitest.config.unit.ts
+++ b/packages/server/vitest.config.unit.ts
@@ -5,6 +5,7 @@ import baseConfig from "./vitest.config.base";
 export default defineConfig({
   ...baseConfig,
   test: {
+    ...baseConfig.test,
     include: ["./tests/unit/**/*.test.ts", "./src/**/*.test.ts"],
   },
 });


packages/server/src/services/mail/templates/raw/reset.html
```diff
@@ -53,7 +54,7 @@
 			font-weight: 500;
 			cursor: pointer;
 			user-select: none;
-			background-color: #484d7c;
+			background-color: var(--site-color, #484d7c);
```
mittalyashu: @coderabbitai Will adding css var will working while rendering/displaying the email in email clients.

mittalyashu: Resolved, thanks @coderabbitai

packages/server/src/services/mail/templates/raw/reset.html
mittalyashu, line 1: @coderabbitai How do you recommend to add test cases for testing the flow sending transactional email.

With assertions for checking the contents of the email is matching as per the injected dynamic data.

mittalyashu: Unit test for `generateContent` seems okay.

For the integrations test, we already spin up a mail service in CI pipeline while running the integrations tests. Can't we use that.

@coderabbitai

mittalyashu: @coderabbit this approach looks good to me.

Create a new PR for this, after I merge this PR.

mittalyashu:
LGTM

PR #1610, opened 2026-03-04
fix(theme): encode dynamic URL path segments with encodeURIComponent

mittalyashu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently encode URL path segments (boards, posts/slugs, roadmaps, roles, API IDs) to handle special characters in navigation and API calls.
  * Prevent navigation when required identifiers are missing (dashboard dropdowns).
  * Added error handling to user role unassignment to ensure errors are propagated.


mittalyashu:
LGTM

PR #1615, opened 2026-03-04
Fix/del site settings cache on update labs

mittalyashu:
## Summary by CodeRabbit

* **Refactor**
  * Scoped lab settings updates to only persist the comments field instead of replacing the full labs payload, reducing risk of unintended overwrites.
  * Enhanced cache invalidation by performing atomic batch operations for setting and clearing related cache entries, improving consistency and performance.
  * Strengthened type assertions for the persisted labs payload to catch mismatches earlier and increase safety when saving settings.


mittalyashu:
LGTM

mittalyashu:
_Slight miscalculation._

This PR will not longer be part of release v0.8.3.

PR #1652, opened 2026-03-11
feat: dynamically import ee routes

mittalyashu:
## Summary by CodeRabbit

* **New Features**
  * Added a consolidated EE API v1 index router.

* **Refactor**
  * Enterprise routes now load lazily at runtime with automatic fallback to open-source mode.
  * Minor middleware import handling streamlined.

* **Tests**
  * Integration setup conditionally initializes EE routes; test suites gain EE-aware wrappers and a skip utility.
  * Test runner configured to run the new setup before integration tests.

* **Chores**
  * CI labels PRs when EE integration tests are skipped.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ server-integration-test:
 	else \
 		pnpm_cmd="pkgx +nodejs.org@22 +pnpm.io pnpm"; \
 	fi && \
+	SKIP_EE_TESTS="false" \
 	NODE_ENV="testing" \
 	LOGCHIMP_SECRET_KEY="secret-key" \
 	LOGCHIMP_WEB_URL="http://localhost:3000" \


Ja4V8s28Ck:
@mittalyashu Is the PR under progress?

mittalyashu:
I have pushed all the commits, I still need to see what CodeRabbit has to say.

Ja4V8s28Ck:
> I have pushed all the commits, I still need to see what CodeRabbit has to say.

I see the tests are failing, that's why. **Failing tests** are intended?

packages/server/src/routes/v1/index.ts
```diff
@@ -1,36 +1,38 @@
...
+  try {
+    const eeRouter = (await import("../../ee/routes/v1")).default;
+    router.use("/api/v1", eeRouter);
+    logger.info("✓ Enterprise features loaded");
+    eeRoutesLoaded = true;
+  } catch (_) {
+    logger.info("Running in open source mode (EE features not available)");
+    logger.info("Upgrade to higher plan to access boards, roadmaps, and roles features.");
+    logger.info("Go to -> https://logchimp.app")
+  }
```
mittalyashu, line 28: @coderabbitai

21-28: Consider implementing single-flight pattern for EE route initialization.

The current boolean guard eeRoutesLoaded only flips after the awaited import settles (line 28). If initialiseEERoutes() is called again before that promise resolves, the guard check at line 23 will pass again, causing router.use("/api/v1", eeRouter) to register the same router twice. Cache the promise instead of the boolean to ensure single-flight semantics and prevent duplicate mounts.

How do you suggest we handle this?

mittalyashu:
_Rebased the PR_

mittalyashu:
@Ja4V8s28Ck ready for review :)

packages/server/tests/utils/skipEE.ts
```diff
@@ -0,0 +1,12 @@
+import { describe, it } from "vitest";
+
+/**
+ * Conditionally skip entire test suites for EE features when SKIP_EE_TESTS is set
+ */
+export const describeEE =
+  process.env.SKIP_EE_TESTS === "true" ? describe.skip : describe;
```
Ja4V8s28Ck, line 7: Can we initialize `SKIP_EE_TESTS` env variable as `true` in `.env.example` so that contributors are aware of this?

mittalyashu: In which `.env.example` file should this environment variable should be added?

Ja4V8s28Ck: 🫠 Ah we don't have a `.env.example` for integration test. We can add it in the [Makefile](https://github.com/logchimp/logchimp/blob/2d667ce7c79c2c6806aeb941fa6cb9f576bbebdd/Makefile#L8-L20).

mittalyashu: > 🫠 Ah we don't have a `.env.example` for integration test. We can add it in the [Makefile](https://github.com/logchimp/logchimp/blob/2d667ce7c79c2c6806aeb941fa6cb9f576bbebdd/Makefile#L8-L20).

Added.

mittalyashu:
@Ja4V8s28Ck Are we good with this PR?

Ja4V8s28Ck:
> @Ja4V8s28Ck Are we good with this PR?

~Not yet, I found some failing tests in the absence of `EE` module. Investigating those tests.~ Apparently, there are some bugs in the integration test. I will make a fix in another PR. LGTM.

PR #1663, opened 2026-03-15
fix(roadmap): skip sort API call when drag is canceled or dropped on same spot

### LINKED ISSUE #1370: Gracefully handle drag canceled or drag to same spot in Roadmaps page
Frontend tried to makes an API call to sort the Roadmaps even when the Roadmap was dragged to the same spot again.

https://github.com/user-attachments/assets/a3dcfeff-c803-419f-99ff-5c26796ac105

sputnik-mac:
## Summary

Fixes #1370 (also addresses the frontend part of #1369)

When a roadmap item is dragged and dropped back to its original position — or the drag is simply canceled — vuedraggable fires the `@end` event with identical `oldIndex` and `newIndex` values. Previously `initialiseSort` always triggered a `sortRoadmap` API call regardless of whether the item actually moved.

### `packages/theme/src/ee/components/dashboard/roadmap/TabularView.vue`
- `initialiseSort` now receives the vuedraggable end event and returns early when `oldIndex === newIndex`, preventing the unnecessary network request.
- Moved `drag.value = false` to the top of the handler so the drag state is always reset, even on an early return.
- Removed the redundant `drag.value = false` from the `finally` block (no longer needed since it's set unconditionally at the start).

### `packages/theme/src/ee/lib/vuedraggable/types.ts`
- Added `VueDraggableEndEvent` interface (`{ oldIndex: number; newIndex: number }`) to give the end-event parameter a proper type.

## Before / After

| Scenario | Before | After |
|---|---|---|
| Item dragged to a different position | ✅ API call made | ✅ API call made |
| Item dropped on same spot | ❌ Unnecessary API call | ✅ No API call |
| Drag canceled (Escape) | ❌ Unnecessary API call | ✅ No API call |

---

I have read the CLA Document and I hereby sign the CLA

## Release Notes

* **Bug Fixes**
  * Improved drag-and-drop sorting in the roadmap view to skip redundant operations when items are not repositioned, providing a more responsive experience.

### DIFF
--- a/packages/theme/src/ee/components/dashboard/roadmap/TabularView.vue
+++ b/packages/theme/src/ee/components/dashboard/roadmap/TabularView.vue
@@ -51,7 +51,10 @@ import { ref } from "vue";
 import type { ISortRoadmapRequestBody } from "@logchimp/types";
 
 import { useDashboardRoadmaps } from "../../../store/dashboard/roadmaps";
-import type { VueDraggableEvent } from "../../../lib/vuedraggable/types";
+import type {
+  VueDraggableEndEvent,
+  VueDraggableEvent,
+} from "../../../lib/vuedraggable/types";
 import { sortRoadmap } from "../../../modules/roadmaps";
 
 import Table from "../../../../components/ui/Table/Table.vue";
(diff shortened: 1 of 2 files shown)


mittalyashu:
LGTM

PR #1670, opened 2026-03-18
fix(auth): handle missing @everyone role in user creation flow

### LINKED ISSUE #1495: Handle missing `@everyone` role in user creation flow
## Description

The `_assignEveryoneRoleQuery` function in `packages/server/src/services/auth/createUser.ts` currently doesn't handle the case where the `@everyone` role doesn't exist in the database. This can cause runtime errors during new user account creation when `getRole` is undefined.

## Impact

This issue affects the user registration flow and could prevent new users from being created if the `@everyone` role is missing from the database.

## Context

The function queries for the `@everyone` role but doesn't verify the result before accessing `getRole.id`, which will throw an error if the role doesn't exist.

## References

- Pull Request: https://github.com/logchimp/logchimp/pull/1494
- Comment: https://github.com/logchimp/logchimp/pull/1494#discussion_r2524220273
- Requested by: @mittalyashu

sputnik-mac:
## What

Guard `_assignEveryoneRoleQuery` against an undefined role lookup result.

## Why

If the `@everyone` role is missing from the database, `getRole.id` throws a `TypeError` and prevents new users from being created. This was identified in [PR #1494 review](https://github.com/logchimp/logchimp/pull/1494#discussion_r2524220273) by @mittalyashu.

## How

- Check if `getRole` is `undefined` before accessing `.id`
- Log an error with a descriptive message if the role is missing
- Return gracefully instead of crashing the signup flow

The user is still created successfully — only the role assignment is skipped with a logged warning. This allows admins to identify and fix the missing role without breaking registration entirely.

Fixes #1495

---
I have read, agree to, and signed the [Contributor License Agreement](https://cla-assistant.io/logchimp/logchimp).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user-creation error handling when the required default role is missing: the system now stops the operation, prevents incomplete user-role assignments, and surfaces an error for proper handling instead of proceeding with partial database changes.

### DIFF
--- a/packages/server/src/services/auth/createUser.ts
+++ b/packages/server/src/services/auth/createUser.ts
@@ -224,7 +224,7 @@ async function _insertUserQuery({
 // assign '@everyone' role
 async function _assignEveryoneRoleQuery(userId: string) {
   const getRole = await database
-    .select<{ id: string }>("id")
+    .select<{ id: string } | null>("id")
     .from("roles")
     .where({
       name: "@everyone",
@@ -232,11 +232,9 @@ async function _assignEveryoneRoleQuery(userId: string) {
     .first();
 
   if (!getRole) {
-    logger.error({
-      message:
-        "Cannot assign '@everyone' role: role not found in database. Ensure the '@everyone' role exists.",
-    });
-    return;
+    throw new Error(
+      "Cannot assign '@everyone' role: role not found in database. Ensure the '@everyone' role exists."
+    );
   }
 
   await database


packages/server/src/services/auth/createUser.ts
```diff
@@ -231,6 +231,14 @@ async function _assignEveryoneRoleQuery(userId: string) {
     })
```
mittalyashu, line 231: <p class='graphite__hidden'><i>[Re: line +227]</i></p>
@sputnik-mac can update the type as well?

```gt_suggestion
    .select<{ id: string } | null>("id")
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/logchimp/logchimp/1670?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/server/src/services/auth/createUser.ts
```diff
@@ -231,6 +231,14 @@ async function _assignEveryoneRoleQuery(userId: string) {
...
     .first();
 
+  if (!getRole) {
+    logger.error({
+      message:
+        "Cannot assign '@everyone' role: role not found in database. Ensure the '@everyone' role exists.",
+    });
+    return;
+  }
```
mittalyashu, line 238: @sputnik-mac Can you test what would the user flow if we throw error?

sputnik-mac:
Addressed both review comments:

1. Updated the type to `.select<{ id: string } | null>("id")` as suggested
2. Changed from `return` to `throw new Error(...)` — if the @everyone role is missing, the user creation will now fail with a proper error instead of silently continuing without a role. This makes sense since a missing @everyone role indicates a database setup issue that should surface immediately.

packages/server/src/services/auth/createUser.ts
sputnik-mac: Done — the type was already updated in the current diff (`.select<{ id: string } | null>("id")`).

packages/server/src/services/auth/createUser.ts
sputnik-mac: Good question. With `throw`, the error is caught in the `createUser` function's `try/catch` block (line ~157), which:

1. Logs the error: `logger.error({ message: "Failed to create user and assign role", error: err })`
2. Returns `500 SERVER_ERROR` to the client
3. Returns `null` — the user record exists in `users` table but has no role assigned

So the user flow: signup request → user row inserted → role assignment fails → **500 response returned** → user sees a server error. The user record exists but can't log in properly until the `@everyone` role is created and assigned.

This is better than the previous behavior (unhandled `undefined.id` crash) because:
- The error message is clear and actionable for the admin
- The response is clean (not an unhandled exception stack trace)
- The user can retry once the admin creates the missing role

I chose `throw` over silent `return` (original version had `return`) since the user shouldn't appear to have signed up successfully when they're missing a critical role. Let me know if you'd prefer different behavior.

PR #1912, opened 2026-09-13
fix(theme): trigger data fetch via InfiniteScroll only when the data lists are empty

mittalyashu:
## Summary

Update `InfiniteScroll` in dashboards (Roles, Boards, Posts, Roadmap) to trigger data fetch when the respective data lists are empty.

**Actual**

A good way to test this is to load data into the component from state management library (for example Pinia store) so the data persists across routes.

In dashboard, when navigating the routes from `Users -> Roadmaps -> Users -> Roadmaps`.

The `roadmaps` route makes the initial API call, and then the second API call with cursor for loading next batch of paginated results without even scrolling the page.

## Type(s) of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] UI
- [ ] Test(s) only
- [ ] Other (Please mention)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial loading for dashboard boards, posts, comments, roadmaps, roles, and users.
  * Empty lists now trigger data loading immediately instead of waiting for a scroll event.
  * Content appears automatically when dashboard sections load without existing items.


mittalyashu:
@coderabbitai

1. Did I missed any other place where this needs to be updated?
2. Is there other way we can address this issue?

mittalyashu:
> A stronger design is to make the shared component accept an initialLoad or shouldLoadOnMount callback. The component should use that condition for both the explicit mount call and useInfiniteScroll’s canLoadMore callback. Each persistent list can then allow an initial load only when its list is empty.

@coderabbitai Doesn't the `immediateCheck` fulfill that requirement?

Even if additional value is added to InfiniteScroll, the component would still get unmounted on navigating to another page/route.

